PR #3, opened 2025-08-21
add cyberpunk accent (diggy's pr)

CometVoid:
see https://github.com/ProjectOmu/OmuStationDepricated/pull/5/files


Orion-416:
Unless this is a port of a trait from NF, this should be in _Omu directories @CometVoid

Orion-416:
Checked with diggy, its from NF

PR #47, opened 2025-08-26
Add Table Crawling

CerberusWolfie:
## About the PR
I've added crawling from Einstein Engines. You can now crawl under tables as most species; the exceptions being Oni and Yowie because some people (Turbo) wanted the big species to be unable to crawl under tables.

I've also made it so if you are crawling under tables, you cannot combat grab anyone (you can still hold them in a combat grab, but you cannot escalate it up a stage and your combo breaks since you will not be able to continue your grab intents).

Since Rodentia should be getting some resistance changes, cough cough #3537, this shouldn't be harmful to their upsides as much. Their sneaking action has been removed since it conflicts with this system.

## Why / Balance
Crawling is cool and adds more interactions. It also just makes sense. As for the system, it can always be removed with a CVar if it isn't liked (as well as reverted), and the only people affected if it isn't reverted are Rodentia since I removed their action.

## Technical details
This modifies the LayingDownComponent to support a new DataField for whether someone is "CrawlingUnder". This adds to the laying down system to change the draw depth of the player when they are crawling under (no modifications to the collide layer outside of those already done so with the laying down portion). This makes them draw under objects just like SmallMobs.

This touches the pulling system, specifically the grab intent section, to prevent people from grabbing while crawling under tables.

I've added tags and a check within the event for crawling under to prevent specific species that are tagged from crawling under tables.

## Media
image
image
image
image

## Requirements

## TODO
These need to be added or something to prevent people from abusing table crawling to dish out damage and whatnot.
- [ ] Add the spray component similar to Oni while crawling under tables.
- [ ] Make melee damage affected by a coefficient (perhaps halving it) while crawling under tables.

## Breaking changes
This moves the LayingDownComponent back out of Goobstation.Common and back into Shared._White.Standing due to the requirement for Shared.DrawDepth to be able to run. It cannot be standalone when it requires the Shared.DrawDepth for the new interactions.

## Referenced PRs
- https://github.com/Simple-Station/Einstein-Engines/pull/1036/

**Changelog**
:cl: CerberusWolfie, Mnemotechnician
- add: Added the ability to crawl under tables to every species except Oni and Yowie!
- tweak: You can no longer combat grab while crawling under tables.


CerberusWolfie:
It's 100% stale, but I'll finish it Soon:tm:.

PR #112, opened 2025-09-09
Salvage Change

harryhulse12-alt:
[

## About the PR
Re-adds the VGroid and allows PKA to work in space

## Why / Balance
Lets salvage work in space without being useless and gives them back the VGroid

## Technical details
Un-comments VGroid and sets minimum pressure for PKA to 0.0

## Requirements

**Changelog**


CerberusWolfie:
You'll need to run this by @Cygnye and @Sarahon as well as some of the maintainers. Part of the reason VGroid was removed was due to the lag it causes. If you want it, it should probably be on an isolated map where it cannot move and doesn't spawn asteroids around it.

Even then so, that'd just be a separate Lavaland knockoff.

PR #120, opened 2025-09-13
Fix basespecies not having face layer

NotActuallyMarty:
## About the PR
1 line yaml for some reason the guy who ported face markings on goob did not include a face layer in the basespecies probably oversight

also includes goob comments for that pr

## Why / Balance
spotted here idk could probably upstream this but rn idc

## Technical details
1 line yaml

## Media
image

before (same face markings):

image

After:
image

## Requirements

**Changelog**
None player facing minor fix.


NotActuallyMarty:
> If its intended to be inherited by other species as well, check them, and while your fixing layering, can you check why feroxi dont have the alt mask layer (might not have alt belt layer either), even though they are parented to BaseMobSpeciesOrganic

Each is dependent on their own species.yml, if any of them redefine the layering, then layers not present wont apply or will apply in odd spots. So feroxi don't have them cause they redefine their own layer order.

Can probably add it, but is there a reason to if markings that use them aren't available?

CerberusWolfie (approved):

PR #126, opened 2025-09-15
Silicon-Interfacing HUD (RDxAI real)

ThanosDeGraf:
Because we like fun in OmuStation!
## About the PR

Added two items that can be equipped, which lets the user communicate on the binary hivemind channel, as well as speak and understand RoboTalk.

## Why / Balance

Since the Binary channel changes, making it a hivemind, RD's binary encryption key no longer worked. ~~But the implanter still does, soooooooooooooooooo, yep!~~ no implanter because it can't be stolen or really interacted with, so had to make them items instead.

To be serious, there are two factions fighting in this debate. One side doesn't want validhunters and metagamers to access binary, since it were to expose malf AIs too quickly. On the other side, many RD and AI mains have expressed it as their favourite relationship, unique in all of the station. Personally, always putting balance concerns of antagonists in front of fun will not lead to more fun RP. Directly enabling RP opportunities will.
After many requests, its here, the Silicon-Interfacing HUD! 
grafik

Put it on and wabam, yer in!

## Technical details

A couple lines of yaml to add it to the RD locker.
New System/Component combo: `ClothingGrantCollectiveMind`. As it says on the tin, a clothing item with this component can add any amount of hivemind channels to the wearer, and remove em when lost. `defaultChannel` is only updated if its already null on the equipee, in case of stuff like Shadowkin who already have their own hivemind set to default.

## Media

image
image
image

https://github.com/user-attachments/assets/96ff04a2-f0fd-4c57-a510-c7aca2a73c9c

## Requirements

## Breaking changes

None, as its all working on new items, sprites and custom system/component

**Changelog**

:cl: Tuerk
- add: New HUD and Glasses that lets RDs (or a sneaky thief) communicate in the binary hivemind. Found in RDs locker and in their loadout.


ThanosDeGraf:
Now just to figure out how to make implanters science/rd contraband

CerberusWolfie (changes requested):
No. Make this a diagnostic HUD glasses (as in, adding a version of it for RD) and tie the equip event to giving access, and the unequip event to removing access.

As stated before, implants cannot really be useful since they cannot exchange across clones, give no room for people to steal them, and give no usefulness to stealing from lockers.

Having them as a form of equippable-item gives you another thing to steal, another item to wear and choose between, and another level of usefulness to the RD.

Diggy0:
The only way i'll accept this is if AI gets a toggle to completely tune out the RD from listening on binary tbh.

3PonPon3:
Or just



Leave it as an implant

Iirc cyg wanted it anywya

ThanosDeGraf:
> Or just
> 
> Leave it as an implant
> 
> Iirc cyg wanted it anywya

Honestly, that's how we'll do it. Later when I or someone else gets a breakthrough, we can replace the implanter with a stealable item.

Diggy0:
Keep it to one. Literally 0 reason they should have two in their locker.

CerberusWolfie:
> Uuuuuuuuuuuuuuuuuuuuuuuuugh this just sucks all around. Every player entity already has a `CollectiveMindComponent`, and trying to add a new one does nothing through the `ClothingGrantComponent` because it will just get skipped, and we dont want to replace it wholesale because some channels could already be present like kinmind or whatever. So somehow I need to check if the comp being added is a CollectiveMindComponent, it already being present, then adding the binary channel to the existing mind component. Then do the reverse for the removing of the clothing...

Why not add a function for adding/removing individual accesses to collective minds? This shouldn't be hard to do, per se.

CerberusWolfie:
Have an event called similar to the clothing grant component (or mimic said component in structure for how its events are called) and have the collective minds added/removed when the item is equipped/unequipped.

ThanosDeGraf:
@CerberusWolfie WABAM! Check it out again, I got your item as promised B)

CerberusWolfie (changes requested):
Make it a pair of sunglasses for RD. Just take the basic color scheme of the diagnostic huds and color that onto some sunglasses similar to how medical sunglasses work. Heads should have sunglasses since they have flashes.

CerberusWolfie:
You can still have the HUD if you want, but the sunglasses should be one as well. If you're worried about it, make the sunglasses part of the spawn for RD and then add both the HUD and sunglasses to the loadout menu for RD.

ThanosDeGraf:
> Make it a pair of sunglasses for RD. [...] Heads should have sunglasses since they have flashes.

Should I add a crafting recipe for the sci sunglasses too, even if the huds are non-craftable?

> You can still have the HUD if you want, but the sunglasses should be one as well. If you're worried about it, make the sunglasses part of the spawn for RD and then add both the HUD and sunglasses to the loadout menu for RD.

When you mean part of the spawn, do you mean a locker item?

ThanosDeGraf:
image
yes

CerberusWolfie (changes requested):

ThanosDeGraf:
Tested and still works yippieee

Content.Shared/_Omu/Clothing/Components/ClothingGrantCollectiveMindComponent.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Content.Shared._Omu.Clothing.Components
+{
+    [RegisterComponent]
+    public sealed partial class ClothingGrantCollectiveMindComponent : Component
+    {
+        [DataField("minds", required: true)]
+        public List<ProtoId<CollectiveMindPrototype>> Minds = new();
+        [DataField]
+        public ProtoId<CollectiveMindPrototype>? defaultChannel = null;
```
CerberusWolfie, line 12:
```suggestion
        [DataField("minds", required: true)]
        public List<ProtoId<CollectiveMindPrototype>> Minds = new();
        
        [DataField]
        public ProtoId<CollectiveMindPrototype>? defaultChannel = null;
```

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/hud.yml
```diff
@@ -0,0 +1,31 @@
+- type: entity
+  parent: [ClothingEyesBase, BaseScienceContraband]
```
CerberusWolfie, line 2: Parent from the original diagnostic HUD so you can just add on the changes, change the sprite, and not need to specify the health bars and whatnot. This is better in design so that changes to diagnostic HUD are relayed to this HUD without having to manually change it (assuming you're not overwriting it with values here).

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/hud.yml
```diff
@@ -0,0 +1,31 @@
...
+  - type: Clothing
+    sprite: _Omu/Clothing/Eyes/scidiag.rsi
+  - type: Tag
+    tags:
+    - HudDiagnostic
+  - type: ShowHealthBars
+    damageContainers:
+    - Inorganic
+    - Silicon
+    - SiliconRadiation
```
CerberusWolfie, line 18: Specifically that these would not be needed since it parents from the Diagnostic HUD.

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/glasses.yml
```diff
@@ -0,0 +1,27 @@
+﻿- type: entity
+  parent: [ClothingEyesBase, ClothingEyesGlassesSunglasses]
```
CerberusWolfie, line 2: And have this parent from the Diagnostic HUD for RD that you made, and it won't need to change anything else except having parented from the sunglasses and changed the sprite.

Resources/Prototypes/Loadouts/role_loadouts.yml
```diff
@@ -525,13 +525,14 @@
...
   - ResearchDirectorHead
+  - ResearchDirectorEyewear # Omustation
   - ResearchDirectorNeck
   - ResearchDirectorJumpsuit
   - ScientistBackpack
   - ResearchDirectorOuterClothing
   - ScientistGloves
   - ResearchDirectorShoes
-  - Glasses
+  # - Glasses
```
CerberusWolfie, line 535: Comment why, don't forget.

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/glasses.yml
ThanosDeGraf: Applied the suggestion and tested, the glasses seem to work still!

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/hud.yml
ThanosDeGraf: Removed!

Resources/Prototypes/_Omu/Entities/Clothing/Eyes/hud.yml
ThanosDeGraf: Done and dusted.

CerberusWolfie (approved):

CliveOcelot (approved):
Tested in game, works

ThanosDeGraf:
kang-council

PR #151, opened 2025-09-26
Allow AI turret controls to set a proper whitelist

CliveOcelot:
## About the PR
Someone made it so that the turret control could only whitelist borg access?, this makes it so that it can whitelist other access, but still defaults to silicon only.

## Why / Balance
Allows AI to clearly define who they want near their core.

## Technical details
Comments out some yml

## Requirements

**Changelog**
:cl:
- tweak: Allowed AI to set whitelists on their turrets properly.

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Wallmounts/turret_controls.yml
+++ b/Resources/Prototypes/Entities/Structures/Wallmounts/turret_controls.yml
@@ -187,12 +187,12 @@
     exemptAccessLevels:
     - BasicSilicon
     - Borg
-  - type: DeployableTurretController
-    accessGroups:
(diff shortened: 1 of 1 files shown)


cheetah1984:
I am pretty sure that upstream made it this way so that AI's den isnt just a safespace for command but is actually just for AI and borgs.

Thesia000:
I need to agree with cheetah here. The AI sat is supposed to be a dangerous place for anything not the AI or a borg. 

The IC reasoning is that NT knows about syndicate operatives and prefers there AI to not be easily attacked,

CerberusWolfie (approved):

PR #170, opened 2025-10-03
Remove Other Rule Entries

CerberusWolfie:
## About the PR
This just adds `ruleEntry: true` to the prototypes for the guidebook entries. Since they are not locked as the default ruleset, they will not show up due to this parameter.

## Why / Balance
Need I explain why?

## Requirements

**Changelog**
Not player facing.

### DIFF
--- a/Resources/Prototypes/Guidebook/rules.yml
+++ b/Resources/Prototypes/Guidebook/rules.yml
@@ -18,9 +18,11 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+
 - type: guideEntry # Default for forks and stuff. Should not be listed anywhere if the server is using a custom ruleset.
   id: DefaultRuleset
   name: guide-entry-rules
+  ruleEntry: true # Omu - removes from guidebook without removing it entirely.
   priority: 60
   text: "/ServerInfo/Guidebook/ServerRules/DefaultRules.xml"
   children:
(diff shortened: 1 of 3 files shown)


MarkerWicker (changes requested):
image

This doesn't seem to have removed the Goobstation rules from the guidebook. The first "Server Rules" tab seen on the left is for Omu, and the second is for Goob.

What you should do, is change the value of the RulesFile CCVar to "OmuRuleset". That should achieve what you're trying to do, and it wouldn't take any changes to YAML, just modifying the server TOML file.

It would also change the rules that're shown in the escape menu when you click on the "Rules" tab.

Check out `CCVars.Server.cs` for reference.

MarkerWicker (approved):
Wait, nevermind, I'm dumb. 
The server does have that cvar set, just not the dev environment. In that case this fix works just fine, as I tested it with the correct cvar value on the dev environment.

CliveOcelot (approved):
Tested in game, seems to work

CerberusWolfie:
> image
> 
> This doesn't seem to have removed the Goobstation rules from the guidebook. The first "Server Rules" tab seen on the left is for Omu, and the second is for Goob.
> 
> What you should do, is change the value of the RulesFile CCVar to "OmuRuleset". That should achieve what you're trying to do, and it wouldn't take any changes to YAML, just modifying the server TOML file.
> 
> It would also change the rules that're shown in the escape menu when you click on the "Rules" tab.
> 
> Check out `CCVars.Server.cs` for reference.

Nope. It does. You have to change your CVAR on your dev environment and it is gone.

PR #172, opened 2025-10-03
Update flavour text limit

Powsu:
## About the PR

Modified default description flavor text limit to 1500
## Why / Balance

Because this limit was originally set by goob, since we are an MRP server, there's a higher expectation of roleplay, and therefore a higher expectation of flavor text. (Also, 512 was too low of a maximum for the lore I was trying to insert into my character.)
## Technical details

I changed a value. That's it. Really.
## Media

image
[added AAA's to show maximum size]

## Requirements

## Breaking changes

If other servers have a lower limit, importing a character save file will cut it off. I "tested" it to avoid making a PR. Alas, my clever efforts were in vain.
**Changelog**
:cl: tweak: Modified maximum flavor text limit. You can now have a bigger character description

### DIFF
--- a/Content.Shared/CCVar/CCVars.Ic.cs
+++ b/Content.Shared/CCVar/CCVars.Ic.cs
@@ -37,7 +37,7 @@ public sealed partial class CCVars
     ///     Sets the maximum length for flavor text (character descriptions).
     /// </summary>
     public static readonly CVarDef<int> MaxFlavorTextLength =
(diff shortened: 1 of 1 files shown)


CerberusWolfie:
I already raised it to 1024 when I enabled them a couple of days ago. We also don't need to hardcode something that can be raised with a CVar.

MarkerWicker:
> 512 was too low of a maximum for the lore I was trying to insert into my character

Isn't the flavour text supposed to be a physical description of the character anyways? Lore can always be explained with words and through the way your character interacts with the world, it's more interesting that way.

PR #262, opened 2025-10-28
Mime muted (paper unlocked)

Lamission877:
## About the PR

-Added a bool that bypasses the speech check when muting someone when true. Added it to the OnCompInit for the mime. 
-Turned off Goob's Mime Enforcement punishments and the illiterate component, so mime can now write if needed. Breaking Vow has become much less punishing if direct communication is needed.

## Why / Balance

Mimes talking in Sign completely bypasses the point of being a mime in my opinion. If you want to communicate normally, just break your vow (The consequences have been turned off as noone wants to be round removed if they just want to talk.) 

After being convinced, Mimes writing on paper can be funny enough so I also turned that off.
## Technical details

I added a variable in MutedComponent that bypasses the RequiresSpeech check when true, so you can fully mute someone. Added a new check to the OnSpeakAttempt method in MutingSystem. Added 2 new argument changes to MimePowersSystem that set the variable to true; one on Init and one on RetakeVows.
Set the PreventWriting bool to false and the PunishmentChance to 0.0 in the MimesPowerComponent.

## Requirements

## Breaking changes

**Changelog**

 :cl:Gab
-tweak: Mime's Vow now blocks Sign Language as well, but the vow no longer prevents you from writing, and you do not get punished for breaking the vow anymore.


Content.Server/Abilities/Mime/MimePowersSystem.cs
```diff
@@ -335,7 +336,8 @@ public void RetakeVow(EntityUid uid, MimePowersComponent? mimePowers = null)
             mimePowers.Enabled = true;
             mimePowers.ReadyToRepent = false;
             mimePowers.VowBroken = false;
-            AddComp<MutedComponent>(uid);
+            var muted = AddComp<MutedComponent>(uid);
+            muted.IgnoreSpeechCheck = true;
```
RichardBlonski:
```suggestion
            var muted = AddComp<MutedComponent>(uid); // Omu
            muted.IgnoreSpeechCheck = true; // Omu
```

RichardBlonski (approved):
Forgot comments

RichardBlonski:
> You should also never use `AddComp`. `EnsureComp` is better and runs a check before attempting to add a component.

Good eye, I haven't noticed those

Lamission877:
> As a Mime main, I disagree. Make it require complex interaction so that they can't talk while holding things.

That's an entirely different thing though, that wouldn't involve the mime at all. I was thinking about doing this, but was struggling with grasping how because the way things work is still quite alien to me. Not sure if I can do that for this PR.

On the subject of reason, I think it's perfectly fine for mime to not be able to speak Sign Language, as that's a method of communication using words, which is against what mime's gimmick should be. Even writing on a piece of paper is not available to a mime, so I think SL should go too.

Lamission877:
Oh my god I dislike existing

Lamission877:
@CerberusWolfie @RichardBlonski Help?

Lamission877:
@CerberusWolfie I added some alleviations for the Sign Language muting from the vow.

Lamission877:
What does this test failure error even mean

Lamission877:
I will be turning it into draft to redo.

CerberusWolfie:
> I will be turning it into draft to redo.

If you're still doing this, the merge from upstream removed mime downsides. You could just make it so you can't use sign language as a mime. Not worried about the paper aspect unless you want to.

PR #275, opened 2025-11-02
Paradox clone disabling until rework

Diggy0:
## About the PR
Disabling Paradox Clones.

## Why / Balance
Okay, long rant time.
Before we received Wizden's version, pretty sure we had DeltaV's back in Goob, which actually allowed for some RP and had both the objectives to co-exist or to kill the original, you wouldn't blow up at CentComm either in round end, etc.,

But then we got Wizden's version, which is extremely closed, and is just generally bad for RP, it's just "Kill this guy and get rid of them" - No RP involved at all, just greentext.
Paradox Clones have had a long history of being instantly defused by the average "what did we do earlier on the shift" question, or a lore question from a friend of yours that knows you ooc etc etc, which is generally unfun for the person playing the paradox.

Admins have also started to reinforce that paradoxes should ALWAYS kill the original, which I find to be dumb, considering the antag itself is just awful and extremely limited in what it can do, it's just walk up to the original, kill, and then try to pretend you are them, which most don't anyways and just play them as their own character but with a different skin.

I'm sick of seeing paradox clones do the most NITRP bullshit, they also lack time requirements so anyone can take them, most paradoxes play out like they were a raider.

## Media
Insert image of paradox clone with "I HATE YOU" text or smth like Richard's PR.

## Requirements
- [X] I am malding after seeing the same exact shit happen over and over with paradox clones

## Breaking changes

**Changelog**
:cl: Diggy
- remove: Paradox clones removed pending rework :godo:

### DIFF
--- a/Resources/Prototypes/GameRules/events.yml
+++ b/Resources/Prototypes/GameRules/events.yml
@@ -186,7 +186,7 @@
     - id: DragonSpawn
     - id: KingRatMigration
     - id: NinjaSpawn
-    - id: ParadoxCloneSpawn
+    #- id: ParadoxCloneSpawn # Omu - Removed till rework, more explanation on the corresponding PR.
     #- id: RevenantSpawn  # GoobStation - removed till rework
     - id: SleeperAgents
     - id: ZombieOutbreak


CerberusWolfie:
Rework it instead of just disabling it. Maybe a little pot calling kettle black, since I disabled Revs, but revs are already being reworked upstream.

Aside from that, Goob never had Paradox Clone until the Wizden pull. We didn't have Delta-V at all. Someone was working on porting it, but decided not to since Wizden already had a system set up.

Also, your Paradox Clone should have to kill you. It makes zero sense why something that is a **_Paradox_** would want to coexist. If you want more RP out of it, then come up with other objectives that it could randomly get instead, and go from there.

ThanosDeGraf:
> If you want more RP out of it, then come up with other objectives that it could randomly get instead, and go from there.

- THE UNIVERSE WILL EXPLODE IF WE COEXIST!!
**Kill your other version before arriving at centcom.** Failure to do so will make both you and your other violently explode from the universe correcting itself.

- I NEED TO WARN THEM OF IMPENDING DOOM!!
A great calamaty befell you, but by a stroke of luck or effort, you managed to go back in time. **Make sure your past self survives until arriving at centcom.**. Once your mission is complete, you will be pulled back into your original timeline.

- IF THAT ANDROID LIVES, IT WILL BE THE END OF ALL OF US!!
Your timeline fell into shambles by a robot uprising, and you were sent back in time to stop their soon-to-be leader. *Make sure this Cyborg/IPC is dead until arriving in centcom.*.
(Additionally, when this objective appears it could spawn Cyborg ghost roles with the objective to protect that person until the paradox is dead)

Honestly there should be enough "I travelled back in time to do X" movies and stories to come up with stronger objectives than these x.x

PR #279, opened 2025-11-03
Wheelchair sprite fixes

ThanosDeGraf:
## About the PR

Improved Vehicle sprite overlay logic. And adjusted wheelchair sprite renderOver directions.

## Why / Balance

Initially I wanted to fix buckled people being unable to interact with anything, but on the localhost that bug doesn't exists. 

So while I was on that, I tried to improve how wheelchairs look when changing directions. Before, it would very frequently struggle with deciding which sprite should be on top. This one, at least in my private testing, seems a lot less erratic, but I would like some more testing.

## Technical details

`VehicleSystem.cs` would use `.GetCardinalDir()`, which sucked balls because it only returned the cardinal directions, where as the `renderOver` attribute accepted the mixed directions as well.

## Requirements

**Changelog**
:cl: Tuerk
- tweak: Wheelchairs should look a little less weird when going in some directions.
- tweak: Wheelchairs are a teeny bit faster on the pedal, still slower than running. Base 'walking' speed unaffected to maintain scenic walks around the station~


CliveOcelot (approved):
looks good to me

PR #284, opened 2025-11-06
[Bounty] Blacklisted the thieving and frisky traits from security and command

TheSecondLord:
## About the PR

Bounty request from https://discord.com/channels/1406768242843979816/1435422233840586854

Characters in security and command positions are now blacklisted from having the thieving or frisky traits. To clarify, having these traits blacklisted doesn't mean you cannot ready up for these positions as a character with these traits; instead, you just won't get the benefit of them while playing these roles. This is noted in the description of both traits.
## Why / Balance

"Only law abiding citizens get hired to NTs finest"
Thieving is a lightly antagonistic trait, and it feels wrong for members of command and security to be able to act that way.
Frisky on security and command is actively against the design philosophy of making pockets usually hidden. Security shouldn't be able to see and report contraband items that the player thought was hidden from sight.

## Technical details

Like 5 lines of yaml (*and 2 lines of ftl) added

## Requirements

## Breaking changes

Nope

**Changelog**

:cl:
- tweak: Members of security and command can no longer have the benefits of the thieving or frisky traits

### DIFF
--- a/Resources/Locale/en-US/_Omu/traits/traits.ftl
+++ b/Resources/Locale/en-US/_Omu/traits/traits.ftl
@@ -54,8 +54,10 @@ trait-thieving-name = Thieving
 trait-thieving-desc =
     You are deft with your hands, and talented at convincing people of their belongings.
     You can steal items more quietly and quickly.
+    Members of security and command will usually lose their ability to do this.
 
 trait-frisky-name = Frisky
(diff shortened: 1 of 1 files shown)


Sarahon:
Just remove frisky and thieving. They are both not good balance and are just a pain.

PR #319, opened 2025-11-23
[Bounty/Feature] Changed Emotespam Smite into a Tempmute

ThanosDeGraf:
## About the PR

Instead of smiting a person that's emote spamming, they will instead be muted from voiced emotes.

### Big Edit: 

This mute is limited to 10 seconds, is limited to just muting the sound of your emotes, allows 5 rapid emotes where every sound emote after has a 16% stacking chance to apply this effect, and at the 10th voiced emote automatically applies the effect.

## Why / Balance

It absolutely fucking sucked when people kill themselves in lightning as soon as they arrive in CC. Noone who is trying to maintain RP after shiftend wants it.

## Technical details

Only changes to goob code, but it's all commented. One new omu status effect.

## Requirements

**Changelog**

:cl: Tuerk
- add: Emote spam causes your emotes to be muted for some time.
- remove: Smite on emote spam removed.


Content.Goobstation.Server/MisandryBox/CatEmoteSpamCountermeasureSystem.cs
```diff
@@ -91,6 +103,18 @@ public override void Update(float frameTime)
...
+        // Omu edit start
+        if (args.Handled)
+            return;
+
+        if (_statusEffectNew.HasEffectComp<EmoteMutedStatusEffectComponent>(ent.Owner)
+            && args.Emote.Category.HasFlag(EmoteCategory.Vocal)) //still leaves the text so it looks like they are pantomiming a laugh
+
+        {
+            args.Handled = true;
+            Log.Debug($"Stfu Entity {ent.Owner.Id}.");
```
RichardBlonski: do we need stfu entity?

ThanosDeGraf: Debug thing

Resources/Prototypes/_Omu/status_effects.yml
```diff
@@ -0,0 +1,11 @@
...
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+# Stops Emotes
+- type: entity
+  parent: MobStatusEffectBase
+  id: StatusEffectEmoteMuted
+  name: emote muted
+  components:
+  - type: EmoteMutedStatusEffect
```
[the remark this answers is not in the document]
ThanosDeGraf: This is an _excellent_ way of doing it because that's what the new status effect system is for. The status effects makes sure to remove itself after the time I specified has ran out, so no unneaded update override needed. And it only activates when someone spams emotes excessively. Introducing a global cooldown is WAY too overkill.

Further, I specifically made it a new effect instead of mime mute because you cannot differentiate between a temporary mime mute and a permanent mute from some chemical.

And I *also* specifcally made it only mutes emotes that make noises as to not stop someone from talking.

ThanosDeGraf:
> Generally a well-loved (hated?) feature

Hated enough for a senior admin to bounty this.

ThanosDeGraf:
> Is there a reason we don't just have a delay after using emote before you can use another? I don't see why muting is a better solution then mechanically fixing the issue. Even a 1-2 second cooldown would eliminate most spam issues.

Because sammy and I dgaf about spamming, it's replacing the smite with something else we're about.

RichardBlonski (approved):
Works ✨

PR #352, opened 2025-12-03
Revert "Revert "bso lifeline to CC""

CliveOcelot:
Reverts ProjectOmu/OmuStation#351

wasnt the PR causing the issue


charcuteriesampler:
testing in prod real

PR #355, opened 2025-12-04
Shadowkin dark adaptation

raincally:
## About the PR

Shadowkin only "Dark adaptation" trait, Burn in light but passive regen in the dark
## Why / Balance

I just want a big gameplay change for shadowkin,,, 

You take 5 burn a second when your light identifier reaches 0
You only heal when your light identifier is max (80) It stops the second you leave the shadows
Your light identifier drains faster the brighter the lights are. You should have enough time to run from department to department before you burn,,, unless someone decides to replaces all the lights with external tubes

Regen doesnt outheal pressure damage. Doesn't heal very much bloodloss so people don't go jumping out of crit all of the time

Let me know if this balancing is shit I can change it,,,
## Technical details

I just copied the shadowling lightdetection code and changed the name and the damage/healing numbers
## Media

https://github.com/user-attachments/assets/45dfcd9b-2c94-4964-b460-5f42d15ffa6d

## Requirements

## Breaking changes

**Changelog**

:cl: raincall
- add: Dark Adaptation trait for shadowkin added

### DIFF
--- a/Content.Omu.Client/LightDetection/ShadowkinLightDetectionSystem.cs
+++ b/Content.Omu.Client/LightDetection/ShadowkinLightDetectionSystem.cs
@@ -1,5 +1,0 @@
-﻿using Content.Omu.Shared.ShadowkinLightDetection.Systems;
-
-namespace Content.Omu.Client.ShadowkinLightDetection;
-
-public sealed class ShadowkinLightDetectionSystem : SharedShadowkinLightDetectionSystem;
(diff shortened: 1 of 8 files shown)


RocketBoss360:
holy peak, im going to ignite
Edit : Actual question, Does this have the potential to ash someone?

dootythefrooty:
True enough, but as it stands it's going to causes issues as well.

RichardBlonski (dismissed):
But yes as suggested by other commenters.

**Shadowlinq** has already a system in-place and components to handle this.

You can grab those components and edit them for when the trait gets added with YAML.
```yaml
- type: ShadowLingComp
  value: X
  value2: Y
```

If you roll Shadowlinq with this trait, guess free buff for being a real Shadowkin

NotActuallyMarty:
> > Please don't directly copy systems when you can just modify the comp values...
> 
> Using shadowling's comp is a bad idea tbh, you then get issues when a shadowkin with this trait becomes a shadowling.

Then make a new comp, There is no need to copy-paste entire systems and have duplicates. Copypasting and renaming a whole system is not the way to do this.

NotActuallyMarty (changes requested):
What was said, why are we copy-pasting an existing system and then just renaming it. Just use the parts that you need and if you need something new make it a dependency. You have things like `LightDetectionComponent` that you can use.

PR #368, opened 2025-12-07
Disable xeno gibing

CliveOcelot:
## About the PR
Chestbursters no longer gib, instead dealing 120 blunt and 80 peirce to the torso, killing the person, but not making them unrevivable.

## Why / Balance
Was requested by a few admins.

## Media
image

## Requirements

**Changelog**
:cl:
- tweak: Chestbursters no longer gib, instead they deal large amounts of damage to the torso.

### DIFF
--- a/Content.Server/_White/Xenomorphs/Larva/XenomorphLarvaSystem.cs
+++ b/Content.Server/_White/Xenomorphs/Larva/XenomorphLarvaSystem.cs
@@ -11,6 +11,8 @@ using Content.Shared.Popups;
 using Robust.Server.Containers;
 using Robust.Shared.Containers;
 using Robust.Shared.Player;
+using Content.Shared.Damage; // Omu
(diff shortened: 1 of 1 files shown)


RichardBlonski (commented):
Good looks good, deals the damage to the chest.

But keep in mind, The larva eats the host from the Inside they should damage or even remove organs when they burst.

Doesn't need to be all of them maybe like the stomach (as you can live without a stomach you just can't drink or eat.)

CliveOcelot:
> Good looks good, deals the damage to the chest.
> 
> But keep in mind, The larva eats the host from the Inside they should damage or even remove organs when they burst.
> 
> Doesn't need to be all of them maybe like the stomach (as you can live without a stomach you just can't drink or eat.)

That's a bit too complex for this PR, and can lead to a decent amount of issues.

RichardBlonski (approved):

Cygnye (approved):
degoidaing

PR #375, opened 2025-12-09
Added A Global DoAfter Delay to All Oni & Predicted the Delay

RichardBlonski:
## About the PR
This PR Adds a Global DoAfter to all Oni. (Makes it 2x longer)

## Why / Balance
Oni are large and it should take them longer to preform actions (This make its 2x longer to do anything with DoAfter)
Also Bounty.

## Technical details
- ~~WoundMed~~ ShitMed already had a GlobalDoAfterMultiplyer which was changed to be predicted (by just making the component be sent to the client, otherwise it will show the normal progress than switch onto the one from the server making the progress bar look unsmooth.
- **Surgical Gloves**, if an Oni uses surgical gloves or tools their Delay will be applied along their multiplyer, so by this PR Oni multiplyer is 2x while surgical gloves is 0.6x, it means with surgical gloves Oni Do After will be 1.2x longer than normal.

## Requirements

**Changelog**

:cl: Richard Blonski
- tweak: Tweak: Oni take longer to preform actions, due to their size.

### DIFF
--- a/Content.Goobstation.Common/DoAfter/DoAfterDelayMultiplierComponent.cs
+++ b/Content.Goobstation.Common/DoAfter/DoAfterDelayMultiplierComponent.cs
@@ -6,14 +6,13 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-using Robust.Shared.GameObjects;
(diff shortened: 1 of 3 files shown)


RichardBlonski:
MISCOMMUNICATION

PR #382, opened 2025-12-12
Fix Spawn and dirty going OOM and failing tests

NotActuallyMarty:
## About the PR

We go OOM due to this bastard 90% of the time. Reverting N14 and disabling pathfinding on it should give us a bit more mem and thus not time us out from workflows.

Merge when build-and-test-debug passes and the other fails are fixed.

## Requirements

## Breaking changes

### DIFF
--- a/Content.IntegrationTests/Tests/EntityTest.cs
+++ b/Content.IntegrationTests/Tests/EntityTest.cs
@@ -57,7 +57,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using System.Text;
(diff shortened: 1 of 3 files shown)


Content.IntegrationTests/Tests/EntityTest.cs
```diff
@@ -57,6 +57,7 @@
 using System.Linq;
 using System.Numerics;
 using System.Text;
+using Content.Omu.Common.CCVar;
```
CliveOcelot:
```suggestion
using Content.Omu.Common.CCVar; // Omu
```

PR #390, opened 2025-12-14
Fix NCT job slots

NotActuallyMarty:
## About the PR

Due to stationjobsystem, it assumes that any infinite job slots are overflow and assigns any overflowing players to those jobs

You can see why this would be problematic if we made a whitelist only role that could overflow.

2 job slots, there should really only be one but idk adminspawn only anyway for now.

Also adds it to blacklist of bureaucraticerror gamerule so that its never set to infinite by that.
## Requirements

## Breaking changes

**Changelog**
Not player facing this was meant to be fixed earlier and changelog is still there


CliveOcelot (approved):
ye

PR #410, opened 2025-12-18
security fire spear

raincally:
## About the PR

okay this might not be a good idea just tell me what u think
dont merge this if its op like the freaking grognard
its a dispatch spear but it does less damage and gives less firestacks,, its also t3 arsenal research
i also gave both the original and lesser dispatch spears actual spear range

if this actually does get merged i mgiht make new sprites for them
## Why / Balance

hopefully sec gets steamrolled by zombies less often
i cant think of any good weapons an oni would be able to use against blob,, terminus isnt great
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl: raincall
- add: Added the lesser dispatch spear, T3 arsenal research

### DIFF
--- a/Resources/Prototypes/Recipes/Lathes/security.yml
+++ b/Resources/Prototypes/Recipes/Lathes/security.yml
@@ -262,10 +262,11 @@
   id: SecDispatchSpear
   result: SecDispatchSpear
   materials:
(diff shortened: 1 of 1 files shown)


WazuraWasTaken:
Baller

PR #430, opened 2025-12-25
Fix Prevent lobby players from hearing barks

maertayn:
## About the PR
When you're in the lobby and doing other things but the game is open in the background you're still able to hear the barks-voices of some clients.

I was unable to reproduce this bug locally with multiple clients, so hopefully this works.
Should be a minimal change that uses AttachedEntity to detect lobby state.

## Why / Balance
Players in the lobby shouldn't hear barks. Audio leaked from the game world into the lobby, which is incredibly annoying when you're just seeding or doing something else.

## Technical details
- Added a simple null check for `AttachedEntity` to detect in-lobby state. It should be null when you're in the lobby or not spawned in, gives an early return to prevent bark audio playing. 
- Worst case bug persists, shouldn't affect players in the game.
- Includes a Copilot suggestion for bark previews when using the lobby's character customisation screen.

## Requirements

**Changelog**

:cl:
- fix: Lobby players should no longer hear barks from in-game players

### DIFF
--- a/Content.Goobstation.Client/Barks/BarkSystem.cs
+++ b/Content.Goobstation.Client/Barks/BarkSystem.cs
@@ -71,7 +71,7 @@ public sealed class BarkSystem : EntitySystem
 
         // Omu inLobby bark-fix
         var session = _playerManager.LocalSession;
-        if (session?.AttachedEntity == null)
+        if (session?.AttachedEntity == null && source != null)
             return;
         // Omu end
 


RichardBlonski:
Gonna keep this till upstrema is merged, because Durk has pushed a fix on Upstream

maertayn:
Their fix, while simple, is much better as it prevents the network event from ever reaching invalid clients in the first place.

PR #447, opened 2025-12-27
new traits

Shy-ivy:
🆑
- add:CPR training and faster surgery traits which medical team and brigmed already know along with some new harpy traits


maertayn:
NT Command hires ought to know CPR as well

PR #100, opened 2025-09-07
[PORT] Impstation Heretic Sacrifice Rework

dootythefrooty:
## About the PR

Ports the Sacrifice Rework from Impstation, which makes it so that sacrifices aren't a total Round Removal. 
## Why / Balance

Heretics being rampant murderboners is fine for a LRP environment but it's questionable for MRP. NLR also means that any sacrificed people shouldn't instantly go for revenge. In the future, it may be prudent to have monsters or Mansus Hands lurking about in the Infinite Archives as a challenge.

## Media

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

**Changelog**

:cl:
- tweak: The Infinite Archives has been ported from Impstation. Heretics sacrificing a target will no longer round remove them. Rather, they will be sent to a book labyrinth briefly before coming back to the station.

PR #500, opened 2026-01-10
[Bounty] Impstation sacrifice rework, and the Infinite Archives

TheSecondLord:
## About the PR

Ported the player sacrifice system from imp. Instead of getting gibbed on sacrifice, you will be transported to the Infinite Archives, a dark maze of bookshelves which exists solely to freak you out a little bit. After ~~15~~ 30 seconds, you will return to the station somewhere in cryo, looking like you've seen a ghost. From the heretic's point of view, it will still look like you were gibbed. NLR should apply upon returning to the station.

This is a revival of https://github.com/ProjectOmu/OmuStation/pull/100 , so credit to dooty for the original port pr.

## Why / Balance

Reduce round removal for MRP, also it's cool and very fitting for heretic's vibes.  

## Technical details

See the original PR. Only a few changes made:
- By default, it tries to respawn you on a midround antag spawn point. It seems that, despite having the component, we don't have markers for these spawn points, nor do we have them mapped anywhere. Instead, it will respawn you on top of a random cryopod in cryo.
- If it can't respawn you at a cryopod, you will remain where you were when sacrificed. Added this so that it doesn't crash if there are 0 cryopods.
- Changed the purgatory time from 15 to 30 seconds because 15 was barely enough to look around.
- Made it so you are very briefly stunned upon being dragged back to the real world.

## Media

https://github.com/user-attachments/assets/b258bdce-aea4-424a-abaa-9ececb9a9e46

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: You will now be temporarily transported to the Infinite Archives upon being sacrified by a heretic, eventually being returned with no memory of what happened.


CerberusWolfie (changes requested):
Develop a system similar to the way TG/Paradise have Heretics where they cannot see the Heretic when sacrificed and will have corrupted organs that will drop.

3PonPon3:
> > Is that really necessary for now....like maybe later sure? But for now I'd say it's fine to be like this to reduce the amount of rr murderbone shit-
> > 
> > (Besides weren't you going to approve the original PR by dooty? It just didn't work in some things?)
> 
> We are not removing RR. Round removal is an innate mechanic in the game. It is intended to exist.

But we're not completely removing RR - we're just removing it from heretic specifically. We can make it be more crippling post the archives or whatever, but like...it is part of the game, but like? Heretic is such an lrp antag gameplay wise already - this serves to make more for an mrp environment. Reference the original reasoning dooty have.

TheSecondLord:
bwomp

Thesia000 (changes requested):
make the teleport random and not based on cryo.

Secondly idealy making the archives random would be grand

PR #551, opened 2026-01-27
Adds BSO surgery Trait

ArtemisYK:
## About the PR
Adds the medical surgery trait to the BSO's yml

## Why / Balance
I find it silly how a command "paramedic" doesn't have SOME form of surgery buff, I didn't want to give them GOD levels of speed increase for power creep reasons so I think the modifier that corpsman and paramedic get fits BSO.

## Technical details
Adds the SurgerySpeedModifier Component to the BSO YML at a 1.25x speed

## Media
Screenshot 2026-01-27 124940

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Surgery Trait to the BSO


Sammyy03:
> BSO is a guard, not a 'paramedic'

BSO's duties are also first-aid (Which doesn't work period due to how healing works for us, there's no "patching up" and then sending you on their way, a paramedic can bring you from crit to near full health with one syringe depending on circumstances) AND **REVIVAL**, "_Their well-being is top priority_". They don't simply act as a body guard, they must ensure the command member/VIP is also healthy at all times, be that by forcing them to go to medbay or healing them on the spot.

This would be a necessity I feel like, considering you cannot revive people with broken bones because you cannot heal them with topicals if they have broken bones. And fixing broken bones is **ungodly** slow.

ArtemisYK:
> > BSO is a guard, not a 'paramedic'
> 
> BSO's duties are also first-aid (Which doesn't work period due to how healing works for us, there's no "patching up" and then sending you on their way, a paramedic can bring you from crit to near full health with one syringe depending on circumstances) AND **REVIVAL**, "_Their well-being is top priority_". They don't simply act as a body guard, they must ensure the command member/VIP is also healthy at all times, be that by forcing them to go to medbay or healing them on the spot.
> 
> This would be a necessity I feel like, considering you cannot revive people with broken bones because you cannot heal them with topicals if they have broken bones. And fixing broken bones is **ungodly** slow.

Put it into words for me perfectly. I was unsure of how to respond without it sounding like i just wanted "tools to powergame" but yeah, you are a guard yes but you are also responsible for the safety and wellbeing of your command. Their safety is top priority and you will not always have time or be ABLE to make it to medbay. Say medical is understaffed, CMO is the person thats dead, YOU are responsible for getting them back up. and with our woundmed, if they are too beaten and bruised they are incapable of being topical'd which... is a need since chems obviously don't work on dead people.

PR #418, opened 2025-12-22
Mary poppins no more! (Warden Rapier is not a jetpack)

Lamission877:
## About the PR
 Added popups to the rapier dash fails (had to also removed autofire from the dash so that the popups don't infinitely appear). made it so that you are not able to dash with the warden's rapier (other dash weapons still dash while weightless though).

## Why / Balance
 Sweet 1500 eggs.

## Technical details
 Added 

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:Gab
- add: popups that show when dash weapons (rogue king's tail, warden's rapier) fail to dash.
- remove: autofire from dash weapons.
- tweak: Warden's rapier can no longer dash while weightless. Go use a jetpack.

PR #635, opened 2026-03-02
Fix wideswing regression caused by PR#418

RadsammyT:
## About the PR
Fixed wideswings being broken on non-dashing AutoAttack weapons as a result of https://github.com/ProjectOmu/OmuStation/pull/418.

## Technical details

The main cause of the bug was this (L158 @ Content.Client/Weapons/Melee/MeleeWeaponSystem.cs):
image

This line to set `wasPreviouslyUsed` did not take into account if the weapon has a `DashingComponent`, which was the main focus for the PR in question. Combined with the fact that it only ran, if the weapon is `AutoAttack` OR no mouse buttons were held down, would make ALL AutoAttack weapons unable to wideswing.

## Media

[Screencast_20260301_231723.webm](https://github.com/user-attachments/assets/8e885f8e-b61e-43c2-a9da-f14995310202)

## Requirements

**Changelog**

:cl:
- fix: Wideswinging on auto-attacking weapons is now fixed.


NotActuallyMarty (approved):
ngl we should probably kill the gauntlets entirely they're cool and all but hella powercreep damn. Or at least yeah kill wideswing for em given they're not a weapon but rather an addition to your fists.

PR #646, opened 2026-03-05
Harpy rework

Holorain4312:
## Why / Balance

This PR makes it so that harpies are faster at using tools that their job requires them to use more often.
It lets medical harpies do surgery faster and and security harpies reload mags with bullet faster.

on the flipside it makes tools that they arent proficient in slower to use

the idea is that this isn't a flat out buff to harpies but a rework.

## Technical details

Added a new prototype for proficiencies

Provided Access to ToolComponent and BallisticAmmoProviderComponent to the system made for this PR

## Media

https://github.com/user-attachments/assets/490558b5-86a3-4079-85d5-c69f49023e9c

## Requirements

## Breaking changes

None to note

**Changelog**
:cl: 
- tweak: Reworked harpies, they are now faster at using tools they are proficient in and slower at using other tools.


Holorain4312:
damn changelog broke it seems

Cygnye (dismissed):

Thesia000 (changes requested):
SO if I am reading this code correct you are setting the proficency GLOBALLY I recomend moving the proficency into the Proicency component instead of having it as a prototype that is saved as a variable inside a system

Thesia000 (changes requested):
for the changes you need to read and save to the entities component not into a variable in the system itself. it will work for local tests but fails once we hit multiplayer as multiple entities will have this comp and override the variable in Proficency System

NotActuallyMarty (changes requested):
ive not looked at the balance i can chud out later at that but you should omumod this instead of putting this in Content.Shared._Omu

Holorain4312:
> Oh it ONLY affects surgert and reloading? For sll harpies I mean
> 
> I thought it would affect things ljkr prying, syringes, using tools, etc etc

So for tools, the way i have it setup:
If the job isnt specified in the yaml, this rework doesnt affect it.
If there is a modifier there, and the tool selected isnt one you are proficient in, you get a negative speed modifier to it; this REALLY comes into effect with med having debuffs to all traditional tools but buffs to surgery.
Also sci will get debuffs for jaws of life experimental welders and drills

NotActuallyMarty (changes requested):
Did some code cleanup, I recommend giving it a once-over and make sure i didn't fuck something up. There were just too many nested statements making code hard to read. Kept functionality the same afaik, so give me a sec to re-review that.

NotActuallyMarty (changes requested):
Math is bad and Prototype refactor.

NotActuallyMarty (commented):
fixing

Holorain4312:
I work on this in a bit

PR #703, opened 2026-03-18
[Bounty] Adds the Shredder

TheSecondLord:
## About the PR

Adds the Shredder, a musician (45tc) and nukie (75tc) only syndicate rock guitar which functions as both a strong melee and as an ally buffing tool. It deals 2 slash/6 blunt unwielded, 4 slash/8 blunt/8 shock (20 total) wielded, has a 1.25x attack rate and grants a buff which gives the wielder 1.2x movespeed and slight passive regeneration of brute, burn and bloodloss damage types while active.
Also adds the blood-red headphones, which allows you to share the Shredder's buff as long as you are within a roughly 4 tile radius of the wielder while worn. 6 of these are included in the nukie version of the listing, so the whole squad can get the buff. They can also be bought on their own for 3tc.

## Why / Balance

Bounty. Provides a very unique form of weapon which will hopefully lead to some interesting nukie strategies. Also, cool job-specific weaponry for syndie musicians.
Healing effect is very minor, so it shouldn't be disastrous for balancing. Values can all be tweaked through yaml if needed.

## Technical details

Added RadiusBuffComponent and RadiusBuffSystem for entities which should apply a status effect to other entities within a radius. Includes fields which should make it flexible enough to be adapted for other uses via yaml in the future, including specifying a different status effect, whitelist/blacklist, if the holder should be affected and if the entity itself should be affected. Somewhat based off of the raw prophet radius heal system, but more generalised. In this case, status effects are applied every 3 seconds, last for 5 seconds and search for entities with the RadiusBuffReceiverComponent (this should really be named ShredderBuffReceiverComponent, oops).

Added the status effect itself, along with ShredderStatusEffectComponent (required by statuseffectsystem and only used to handle the state of the status effect) and ShredderBuffComponent (component given to entities with the status effect which holds most of the datafields). Added EffectShredder proto for the indicator visual. Healing and visual is applied every 3 seconds. Movement speed is granted/removed and refreshed when the status effect is added/removed. 

Added ActivateBuffOnWieldComponent and ActivateBuffOnWieldSystem. Toggles the radius buff on/off when wielded/unwielded. Not really made to be used with anything other than RadiusBuffComponent.

Added uplink listings for the shredder, nukie shredder bundle and bloodred headphones.

## Media

https://github.com/user-attachments/assets/fe169bf1-e945-4b4c-8a90-130374168a1d

Nukie bundle listing
image

Roughly maximum buff range
image

## Requirements

## Breaking changes

**Changelog**

- add: Added the Shredder weapon for syndicate musicians and nuclear operatives. Dance till you're dead.


Thesia000 (changes requested):
the animation is mininmal off please fix it, make sure it is centered on a urist

NotActuallyMarty (changes requested):
These are general notes i've not looked over this entirely, but the calls of EQE on an update loop don't look good to me.

You made a `ShredderStatusEffectComponent` But no actual entity for the Component.

Make an entity prototype in yaml, and probably move the effects from `ShredderBuffComponent` to it, then apply it where you need to when you need to using status effect system.

See for example the status effect for tasers:
```
- type: entity
  parent: StatusEffectSlowdown
  id: TaserSlowdownStatusEffect
  name: shot by taser slowdown
```
with the parent being:
```
- type: entity
  parent: MobStatusEffectDebuff
  id: StatusEffectSlowdown
  abstract: true
  name: slowdown
  components:
  - type: StatusEffect
    whitelist:
      components:
      - MobState
    blacklist:
      tags:
      - SlowImmune
  - type: MovementModStatusEffect
 ```
You are going to probably need to do a bit more given the stuff you want to give, but for example just by adding a `MovementModStatusEffectComponent` to your status effect already simplifies this by a lot because MovementModSystem will handle the movespeed refreshing without you needing to hardcode things.

You could also probably move this to OmuMod which would make it a lot cleaner. Just put the systems in `Content.Omu.Shared` or `Content.Omu.Server` or whatnot, and if you don't need to touch core files that should work.

**Comment by NotActuallyMarty** [comment]
Okay you did add a status effect entity my ass is blind, this just needs some cleanup then

**Comment by TheSecondLord** [comment]
Reviews addressed for the most part. I'm a bit iffy on the EQEs too but it does need to refresh on a timer so I think this is acceptable?

> You could also probably move this to OmuMod which would make it a lot cleaner. Just put the systems in Content.Omu.Shared or Content.Omu.Server or whatnot, and if you don't need to touch core files that should work.

What's actually the difference between this and just namespacing it?

File: Content.Omu.Shared/RadiusBuff/Components/ActivateBuffOnWieldComponent.cs [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+/// Activates a <see cref="RadiusBuffComponent"/> when this entity is wielded
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class ActivateBuffOnWieldComponent : Component
+{
+    /// <summary>
+    /// Deactivate on wield instead
+    /// </summary>
+    [DataField]
+    public bool Invert = false;
```
NotActuallyMarty: default on bool is false
```suggestion
public bool Invert;
```

Content.Omu.Shared/RadiusBuff/Components/RadiusBuffComponent.cs
```diff
@@ -0,0 +1,64 @@
...
+
+/// <summary>
+/// Defines a status effect that an entity should give in a radius under some condition.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+[AutoGenerateComponentState, AutoGenerateComponentPause]
+public sealed partial class RadiusBuffComponent : Component
+{
+    [DataField]
+    public bool Active = false;
```
NotActuallyMarty:
```suggestion
public bool Active;
```

Content.Omu.Shared/RadiusBuff/Components/RadiusBuffComponent.cs
```diff
@@ -0,0 +1,64 @@
...
+    public EntityWhitelist? Whitelist;
+
+    [DataField]
+    public EntityWhitelist? Blacklist;
+
+    /// <summary>
+    /// If this buff should apply to the entity that is providing this buff, regardless of the whitelist.
+    /// </summary>
+    [DataField]
+    public bool BuffSelf = false;
```
NotActuallyMarty:
```suggestion
public bool BuffSelf;
```

Content.Omu.Shared/RadiusBuff/Systems/ShredderStatusEffectSystem.cs
```diff
@@ -0,0 +1,71 @@
...
+            Dirty(uid, comp);
+        }
+    }
+
+    private bool TryApplyHealing(EntityUid target, ShredderStatusEffectComponent comp)
+    {
+        if (comp.HealBuff == null)
+            return false;
+
+        if (!TryComp<DamageableComponent>(target, out var damageable))
```
NotActuallyMarty, line 55: always false since non nullable and you are putting this in the component

```suggestion
    private bool TryApplyHealing(EntityUid target, ShredderStatusEffectComponent comp)
    {
        if (!TryComp<DamageableComponent>(target, out var damageable))
```

TheSecondLord:
controversial opinion the bot shouldn't do this if a PR has been "inactive" because of review hell. should only do this if it has the "awaiting changes" or "merge conflict" tag

Thesia000 (approved):

NotActuallyMarty (approved):

NotActuallyMarty:
You can definitely entityeffect this further i'll do it myself on upstream refactor

PR #737, opened 2026-03-25
Executioner's Axe

Giltsmith:
## About the PR
As per the request of our _magnanimous_ leader, adds a new selectable weapon for the warden: an executioner's axe. Can be raised to use as a shield, slows you while walking/sprinting, only swings while wielded: and it decaps on execution. 

## Why / Balance
Adds a new way for Ward to punish enemies pushing into a backline. Mechanically I think it's cool to have a weapon you can use as a shield, and it shouldn't be overly effective against range. Just... step away. They can't follow you and maintain defense.

## Technical details
All yaml, nothing fancy.

## Media

https://github.com/user-attachments/assets/4f6314bd-0708-47af-9920-4fcf035eed4f

## Requirements

**Changelog**
:cl: Giltsmith
- add: Warden's Axe


CliveOcelot (changes requested):
Either make the shield require charge, or remove it outright, We do not need more shields that are unbreakable

PR #740, opened 2026-03-25
Resomi reworked 2.0: restfulness, chirp directions and nesting toggle

Raze500:
## About the PR
Reworks the Resomi species with 4 new unique mechanics that reinforce their identity as fast, fragile scouts rather than frontline fighters.

Reworks Resomi species mechanics. Replaces the broken v1 systems with properly implemented ones and adds a chirp coordination ability.

**Changes**
Restfulness - a passive bar that drains over time and only recharges while nesting. Low restfulness reduces movement speed. Replaces the old energy system which never actually drained (it depended on StaminaComponent which Resomis do not have).

**Nesting** - immobilizes the Resomi but reduces incoming damage and heals over time. Cannot be used while cuffed. Resomis can still rotate, speak, and chirp from the nest.

**Chirp** - sends a compass direction to nearby Resomis (45 tile range, same map). Shows as a floating notification above the character, not in chat. 1 second delay before receivers get it. Works while cuffed or nesting. Logged for admins

## Why / Balance
Resomis previously had no mechanics that distinguished them from other species in combat. This rework gives them a clear niche: high mobility and support utility, with significant penalties that punish sustained combat engagement.

## Technical details

New reusable system
`PopupEntityNoChatEvent` (Content.Omu.Shared) now shows a popup above an entity in world space without logging to chat, regardless of the `chat.log_in_chat` cvar for any future server system can use it.

## To do:

- [X] Fix the YAML error
- [X] Nerf a little the stamina damage when using weapons.

## Media

https://github.com/user-attachments/assets/a91af261-4bd4-43a6-80b1-08a8698eeb74

https://github.com/user-attachments/assets/90ba5ca1-314d-463e-b574-7e190feeb531

## Requirements

## Breaking changes
None.

**Changelog**
:cl: NEVEC
add: Resomis now have a Nesting ability immobilizes them but reduces damage taken and heals over time
add: Resomis have a Restfulness bar that drains over time and only recovers while nesting; low restfulness reduces movement speed
add: Resomis have a Chirp ability that sends their direction to nearby Resomis as a floating notification. Works while cuffed or nesting
tweak: Resomis deal 15% less melee damage


Raze500:
I will fix this and tweak some things later, meanwhile please if somebody review this tell me if its fine.

ThanosDeGraf (commented):
I like the direction you are giving to Resomi! Having a glance at it, here are some suggestions

CerberusWolfie (changes requested):
The staff are somewhat divided on Nesting, so I will ignore it for now, but this is the general feedback I have.

- The melee debuff is unnecessary and I simply don't want it. There are several reasons for this, but the primary is that it just isn't fun being weaker than the average species. This melee debuff would compound with weak resistances Resomi already have. I don't want to play a species that only makes up for its vast negatives by having healing abilities and other interactions.
  - If you want another reason, one of the arguments for changes in the main channel was that these are raptors and should have unique style; raptors had sharp ass claws intended for piercing, clawing, and gouging their prey (based on the fossil), so that shouldn't be any different here.
- The gun debuff shouldn't be unique to Resomi. It should be a separate system that could be applied to other species in the future; from a maintainer standpoint.
- The gun debuff, in general, would be better tuned to being a knockback on the user firing, based on the direction they are firing. Doing stamina damage is taxing and there is already many factors affecting stamina.
  - Perhaps we could extend it to more stamina damage in the future, but my current point is that it would be better to not affect it.
  - The spraying based on size would be fine, I'm not too worried about it.
- The chirping system would be neat to be more of a subtle message that isn't recorded in chat, with a cooldown. This would promote players to be paying attention and it would be neat to have something that would require attention and immediate action.
  - There is more to this, but this is my idea currently.
- The separate energy system needs to go. We do not need more complexity to a species and especially do not need a niche system that reflects a separate stamina bar.
  - My main reason for this is that it affects the player, forces another mechanic, and divides the purpose of the species and interactions in combat.
  - I will also point to other games, in this case, that do not have separate systems for "stamina" or "energy", being many fighting games and even Skyrim as an example (albeit modding makes this different).

All in all, I do agree that they could use some more uniqueness, but, from a maintainer perspective, a lot of this could be generalized into systems that could be applied to others without always being "resomi-comp". For example, the pseudoitem was for felinids and it is applied to more than that and isn't called "feliniditem".

On top of that, adding more systems that only have one use case and are niche to a species and use nowhere else would have more complexity that isn't entirely needed when we could be utilizing already-existing systems. One example is the melee debuff being entirely coded for Resomi when there already exists the Oni system which could be applied, as mentioned by Tuerk earlier.

Content.Client/Popups/PopupSystem.cs
```diff
@@ -158,7 +159,8 @@ private void PopupMessage(string? message, PopupType type, EntityCoordinates coo
 
             _aliveWorldLabels.Add(popupData, label);
 
-            if (_shouldLogInChat &&
+            if (!skipChatLog &&
```
ThanosDeGraf, line 162: Omu comment here

Content.Omu.Client/Resomi/EntitySystems/NestingFrozenSystem.cs
```diff
@@ -0,0 +1,9 @@
+using Content.Omu.Shared.Resomi.EntitySystems;
+
+namespace Content.Omu.Client.Resomi.EntitySystems;
+
+/// <summary>
+/// Client-side concrete implementation of SharedNestingFrozenSystem.
+/// Required so movement blocking applies during client-side prediction.
+/// </summary>
+public sealed class NestingFrozenSystem : SharedNestingFrozenSystem;
```
ThanosDeGraf, line 9: Since the client and server side systems are not different from each other and likely will never be, you can remove this and rename SharedNestingFrozenSystem to NestingFrozenSystem.

Content.Omu.Server/Resomi/EntitySystems/NestingFrozenSystem.cs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+using Content.Omu.Shared.Resomi.EntitySystems;
+
+namespace Content.Omu.Server.Resomi.EntitySystems;
+
+/// <summary>
+/// Server-side concrete implementation of SharedNestingFrozenSystem.
+/// </summary>
+public sealed class NestingFrozenSystem : SharedNestingFrozenSystem;
```
ThanosDeGraf, line 12: Same here, can be removed.

Content.Omu.Server/ExtraResource/EntitySystems/ExtraResourceSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+using Content.Shared.Movement.Systems;
+
+namespace Content.Omu.Server.ExtraResource.EntitySystems;
+
+/// <summary>
+///     manages all ExtraResourceComponent entries each tick.
+///     currently handles the "restfulness" key for Resomi (and any future species that use it).
+///     to add a new resource for another species, add a new key check in Update.
+/// </summary>
+public sealed class ExtraResourceSystem : EntitySystem
```
ThanosDeGraf, line 17: Shouldn't this be in shared?

Content.Omu.Server/ExtraResource/EntitySystems/ExtraResourceSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    // sprint speed boost when fully rested (above 80%)
+    private const float RestfulSprintBoost = 1.15f;
+
+    // sprint speed penalty when exhausted (below 30%)
+    private const float ExhaustedSprintPenalty = 0.80f;
+
+    // thresholds for the sprint modifier transitions
+    private const float RestfulThreshold = 80f;
+    private const float ExhaustedThreshold = 30f;
```
ThanosDeGraf, line 36: Shouldn't these be moved to the component?

These are too many constants in a generic system, and the names are too resomi specific. I'll wrack my own brain on how to deal with this with you.

Content.Omu.Server/Resomi/EntitySystems/NestingSystem.cs
```diff
@@ -0,0 +1,179 @@
...
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Systems;
+using Content.Shared.Popups;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Player;
+using Robust.Shared.Timing;
+
+namespace Content.Omu.Server.Resomi.EntitySystems;
+
+public sealed class NestingSystem : SharedNestingSystem
```
ThanosDeGraf, line 22: Most if not all of these methods should be moved to shared. The only reason it seems you can't is because your ExtraResource comp and system is not shared.

Only the client system dealing with sprite changes is good.

Content.Omu.Server/Resomi/EntitySystems/NestingSystem.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            // check if a wind-up is pending and has completed
+            if (comp.PendingToggleAt is { } pendingAt && curTime >= pendingAt)
+            {
+                comp.PendingToggleAt = null;
+                if (comp.IsNesting)
+                    DoExitNest(uid, comp);
+                else
+                    DoEnterNest(uid, comp);
+                continue;
```
ThanosDeGraf, line 60: Not sure if you need to skip to the next entry here. But not really that big of a deal.

Content.Omu.Server/Resomi/EntitySystems/NestingSystem.cs
```diff
@@ -0,0 +1,179 @@
...
+    {
+        comp.IsNesting = true;
+        comp.NextHeal = _timing.CurTime;
+
+        EnsureComp<NestingFrozenComponent>(uid);
+        _audio.PlayPvs(comp.NestEnterSound, uid);
+
+        var coords = Transform(uid).Coordinates;
+        Spawn(comp.NestEnterEffect, coords);
+        comp.ContinuousEffectEntity = Spawn(comp.NestContinuousEffect, coords);
```
ThanosDeGraf, line 143: Not sure about manually spawning a whole new entity for this.

With the [New Entity Effect System](https://github.com/space-wizards/space-station-14/pull/37238), you'd achieve a similar effect.

Resources/Prototypes/_Omu/Actions/resomi.yml
```diff
@@ -0,0 +1,27 @@
...
+    icon:
+      sprite: _Omu/Interface/Actions/actions_resomi.rsi
+      state: nest
+  - type: InstantAction
+    event: !type:ToggleNestActionEvent
+
+- type: entity
+  id: ActionResomiCall
+  name: Call
+  description: Let out a piercing raptor call. Other Resomis on the station will hear it and know which direction it came from.
```
ThanosDeGraf, line 19: Does not work as described here, as the call only affects resomi within a internally specified range, not station-wide.

Content.Omu.Server/Resomi/EntitySystems/ResomiCallSystem.cs
```diff
@@ -0,0 +1,189 @@
...
+    private void OnShutdown(EntityUid uid, ResomiCallComponent comp, ComponentShutdown args)
+    {
+        _actions.RemoveAction(uid, comp.CallActionEntity);
+    }
+
+    private void OnCall(EntityUid uid, ResomiCallComponent comp, ResomiCallActionEvent args)
+    {
+        var callerName = MetaData(uid).EntityName;
+        var callerXform = Transform(uid);
+        var callerPos = _transform.GetWorldPosition(callerXform);
```
ThanosDeGraf, line 104: How the call determines direction is "correct", but only works if the station is already rotated aligned to north. 

On the dev map, I had placed one resomi in atmos and another at the start area. When the atmos resomi called, the pop-up for the start-resomi says "Southeast". On the mass scanner, this is correct. 

Image


But on the station map, it's a totally different direction.

Image

Maybe someone else has a different opinion on this, but I think the direction should not be given relative to the world, but relative to the grid the receiving resomi is on.

Content.Omu.Server/Resomi/EntitySystems/ResomiCallSystem.cs
```diff
@@ -0,0 +1,189 @@
...
+    private void SendNoChatPopup(string message, EntityUid recipient, PopupType type)
+    {
+        if (!TryComp(recipient, out ActorComponent? actor))
+            return;
+
+        RaiseNetworkEvent(
+            new PopupEntityNoChatEvent(message, type, GetNetEntity(recipient)),
+            actor.PlayerSession
+        );
+    }
```
ThanosDeGraf, line 157: Why do we not want to log a call to chat?

Content.Omu.Shared/Popups/OmuPopupEvents.cs
```diff
@@ -0,0 +1,56 @@
...
+///             RaiseNetworkEvent(
+///                 new PopupEntityNoChatEvent("your message", PopupType.Large, GetNetEntity(uid)),
+///                 actor.PlayerSession
+///             );
+///
+///     uid   = the entity the popup floats above (usually the same as recipient)
+///     recipient = the player who will see it (actor.PlayerSession)
+/// </remarks>
+[Serializable, NetSerializable]
+public sealed class PopupEntityNoChatEvent : PopupEvent
```
ThanosDeGraf, line 48: Holy mother of doc strings 

Image

Content.Server/Destructible/Thresholds/Behaviors/EmptyAllContainersBehaviour.cs
```diff
@@ -31,7 +31,11 @@ public void Execute(EntityUid owner, DestructibleSystem system, EntityUid? cause
 
             foreach (var container in system.EntityManager.System<SharedContainerSystem>().GetAllContainers(owner, containerManager))
             {
-                system.ContainerSystem.EmptyContainer(container, true, system.EntityManager.GetComponent<TransformComponent>(owner).Coordinates);
+                // do not pass explicit coordinates - AttachParentToContainerOrGrid handles both
+                // grid and space cases correctly. passing owner.Coordinates as destination
+                // fails the assert in SharedContainerSystem.Remove when the entity is in space
+                // and SetCoordinates cannot place the item at the intended position.
+                system.ContainerSystem.EmptyContainer(container, force: true);
```
ThanosDeGraf, line 38: Whats this doing here

Content.Omu.Server/ExtraResource/EntitySystems/ExtraResourceSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    case "restfulness":
+                        UpdateRestfulness(uid, comp, comp.Resources[key], frameTime);
+                        break;
+                }
+            }
+        }
+    }
+
+    private void UpdateRestfulness(EntityUid uid, ExtraResourceComponent comp, ExtraResourceEntry entry, float frameTime)
```
ThanosDeGraf, line 63: Standing still, sprinting, nesting or rejuvenating, I could not get `Current` below 100.

Plus even at maximum restfulness, the alert does not change (I saw it be correctly only once and that was when spawning a fresh resomi.)

Image

ThanosDeGraf (commented):
I should go to sleep, but I should look over this again. More than a couple nitpicks, sorry.

Content.Omu.Server/Resomi/EntitySystems/ResomiCallSystem.cs
Raze500: I remember i did talked about the direction about if i should do it on the map direction and it was fine with it beacuse most of the people uses the default camera

Content.Omu.Server/ExtraResource/EntitySystems/ExtraResourceSystem.cs
Raze500: It usually takes time, i used on my tested version to make it fast and drop to see if it worked, gonna check if its a problem about the server then

Content.Omu.Shared/Popups/OmuPopupEvents.cs
Raze500: in the doc looked more nicier :(

Content.Omu.Server/Resomi/EntitySystems/ResomiCallSystem.cs
ThanosDeGraf: That WAS with the default camera alignment.

Raze500:
To organize myself, i will leave this on draft, gonna focus on ther main project firsts and then fix all this mess!

ThanosDeGraf:
> To organize myself, i will leave this on draft, gonna focus on ther main project firsts and then fix all this mess!

Good going and good luck!

PR #804, opened 2026-04-07
Fixes Thaven CANNOT work with internals

Raze500:
## About the PR
Fixes Thaven not breathing with internals in space.

## Why / Balance
Thavens were still suffocating in vacuum even with valid internals running, which is a major gameplay bug. this change makes their breathing logic respect actual inhaled gas and allows standard low-pressure internals output to count as a valid to breath

## Technical details
- thaven breathing system  now checks whether the Thaven can metabolize the gas they would actually inhale, instead of only checking tile atmosphere 
- Set the minPressure to 5 on standard thaven lungs so standard internals output is treated as breathable.

## Media

https://github.com/user-attachments/assets/521152cc-ce39-458e-9401-ad5c304e28d1

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-pr-guidelines.html).

## Breaking changes
None.

**Changelog**
:cl: Nevec
- fix: Fixed Thavens suffocating in space while using valid internals.

### DIFF
--- a/Content.Server/_Impstation/Thaven/Systems/ThavenBreatherSystem.cs
+++ b/Content.Server/_Impstation/Thaven/Systems/ThavenBreatherSystem.cs
@@ -1,7 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using Content.Goobstation.Shared.Body;
 using Content.Server._Impstation.Thaven.Components;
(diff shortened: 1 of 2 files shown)


Raze500:
ayo what

PR #810, opened 2026-04-07
[TWEAKS/FIXES] Port PR 1745 from funky (Crew manifest changes)

CliveOcelot:
## About the PR
Crew manifest now scales text such that it will never overflow names onto the job section, and also shows gender of the character.

## Why / Balance
Thaven names were causing issues with how long they were, this fixes it.

## Technical details
To note, all cyborgs are marked with their gender set to other, as I could not work out a easy way to pass it through to the Omu system.

## Media
image

## Requirements

**Changelog**
:cl:
- tweak: Crew manifest now shows the gender of crew. 
- fix: Fixed long names overflowing into the job section of the crew manifest.

### DIFF
--- a/Content.Client/CrewManifest/UI/CrewManifestSection.cs
+++ b/Content.Client/CrewManifest/UI/CrewManifestSection.cs
@@ -103,27 +103,59 @@ public sealed class CrewManifestSection : BoxContainer
             Text = Loc.GetString(section.Name)
         });
 
-        var gridContainer = new GridContainer()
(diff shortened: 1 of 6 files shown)


NotActuallyMarty (changes requested):
you would change the job prototype to pass through that.

don't do it i will find you.

probably better to pass it as `neuter` rather than `other`?

NotActuallyMarty (approved):

Thesia000 (approved):

PR #820, opened 2026-04-08
Electropack ID lock system with access control and electric shock

Raze500:
## About the PR
Adds an ID lock system to the electropack. Authorized roles can lock and unlock it via right-click, either on the backpack itself or on the person wearing it. Once locked, nobody without clearance can strip it off - and if they try, they get shocked.

## Why / Balance
The electropack is high-value tactical equipment. Right now anyone can just yank it off someone mid-combat with no consequence. This gives it weight - you have to actually earn access to touch it, and trying to steal it has a real cost.

## Technical details
- Added SelfUnremovableClothingComponent to the electropack with ascess that can read with the next ascess: (Captain, Armory, Corpsman, Blueshield, NT Rep)
- Lock/unlock verb shows up both on the item and when right-clicking the person wearing it
- Unauthorized strip attempts are cancelled on both client and server (prevents the item briefly flashing into the hand) (This was a bug and now is fixed)
- Strip UI spam (60 events/s while holding the button) is throttled with a 2-second cooldown (This was a bug, fixed)
- Electric shock on unauthorized strip is handled server-side via a custom event to avoid SS14's duplicate subscription restriction
- Shock damage and stun duration are DataFields, disabled by default on the component - enabled only on the electropack prototype

## Media

https://github.com/user-attachments/assets/89e8f59d-5c42-425e-8615-9c6f91246b70

https://github.com/user-attachments/assets/409a1cba-af2a-41b7-aae7-0fb02f8fe2c2

https://github.com/user-attachments/assets/3728b005-19f0-4f64-a5e2-3c8c21329931

https://github.com/user-attachments/assets/d14606f1-470f-45d7-85d0-bb34fa58a0de

## Requirements

## Breaking changes
None.

:cl: Nevec
- add: The electropack can now be locked by authorized personnel. Unauthorized strip attempts are blocked and will shock the offender.
- add: You can bypass locking feature with any sharp object


CerberusWolfie (changes requested):
Stop requesting `@Aidenkrz`, he does not want to review these PRs, for sure for sure.

Other topic, as discussed in the Discord, the electropack should be removable via a cutting item (shiv or equivalent) if we're going to implement a system like this.

The person wearing it should be able to remove it as well but with a longer doAfter. You may apply like a 5 shock if they lack insuls, but they should always lack insuls in this case, so I think it is just punishment on top of it, but it's whatever.

You can make it like a 10 second doAfter for another prisoner and a 15 second doAfter for the electropacked user.

NotActuallyMarty (changes requested):
on god kill the vibecode.

Thesia000 (changes requested):
Please change the access to security instead of the 5 subaccesses

PR #824, opened 2026-04-09
Restore wizden dragon

CliveOcelot:
## About the PR
Replaces the version of dragon goob has with the one from upstream wizden, Also adds a new version of ichor to be used by the dragon (same stats as wizden's ichor)

## Why / Balance
Our current dragon is far too strong. And has a incredibly low TTK.

## Technical details
Copies code from wizden, replaces our code with it.

## Requirements

**Changelog**
:cl:
- tweak: Replaced current dragon with Wizden's dragon.


Diggy0:
I've spectated like 3 dragon rounds in Omu (Yes. We had Goob's dragon when I spectated them.) and it's always the same skill issue on sec's side.
They never push and just camp out the dragon like it was gonna get out of the zone they have their carps on, pff.

This whole thing just leads to the dragon piling up a massive amount of Carp, if sefc doesn't push early, they're never gonna win, like, ever.

I'm pretty sure Wizsec is way weaker than Goob / Omu sec, so completely reverting dragon to the "vanilla" state per say is just begging for Sec to curbstomp them in like 0,2 just to go back to nothing ever happens, if security is struggling with a single dragon 99% of the rounds i've seen is because they can't coordinate even if they were held at gunpoint / have the worst protag mindset i've seen in ages, they don't coordinate, they don't plan out, they don't use their resources.

This PR just feels like someone of the staff team lost to dragon / players don't know how to fight dragons.
Reverting dragon to 40 pop requirement after nerfing it to the ground is dumb sob.

Anui-ElXx:
Just remove dragon roar (chud ability) and give it Dragon Kung Fu

Paskenduolis:
Even if omu sec sucks, the dragon is still too powerful and dangerous for a minor antag as defined by the manual. When a minor antag apears, people shouldn't have to rush longarms, get geared up as much as possible, and coordinate all of their manpower to take down one threat. Which is why I think the dragon should either be nerfed or changed into a major antag

Diggy0:
> Even if omu sec sucks, the dragon is still too powerful and dangerous for a minor antag as defined by the manual. When a minor antag apears, people shouldn't have to rush longarms, get geared up as much as possible, and coordinate all of their manpower to take down one threat. Which is why I think the dragon should either be nerfed or changed into a major antag

Dragon is not powerful
Imma be fr the main issue is Omu hates antags with their whole being and god forbid a midround does something other than be quirky and break it down.
Dragon wouldn't win most of the time if omu-sec actually did their job instead of giving up the moment the antag can fight back and it isn't just a one-sided magdump session.

Anui-ElXx:
I was thinking about making a new "Real Dragon Kung Fu" exclusively for Dragon.

It would have all the same moves as dragon king fu (maybe we make it deal a bit more damage because Dragons are stronger than humans?), but it would also have moves that set your opponent on fire or throw a smaller fireball.

I just think it would be peak and make Dragon actually more unique.

Diggy0:
> I was thinking about making a new "Real Dragon Kung Fu" exclusively for Dragon.
> 
> It would have all the same moves as dragon king fu (maybe we make it deal a bit more damage because Dragons are stronger than humans?), but it would also have moves that set your opponent on fire or throw a smaller fireball.
> 
> I just think it would be peak and make Dragon actually more unique.

We don't need more martial arts slop tbh.
Inb4 this is just saying random stuff and praying it lands and I fell for it or whatev.

Anui-ElXx:
> > I was thinking about making a new "Real Dragon Kung Fu" exclusively for Dragon.
> > 
> > It would have all the same moves as dragon king fu (maybe we make it deal a bit more damage because Dragons are stronger than humans?), but it would also have moves that set your opponent on fire or throw a smaller fireball.
> > 
> > I just think it would be peak and make Dragon actually more unique.
> 
> We don't need more martial arts slop tbh.
> Inb4 this is just saying random stuff and praying it lands and I fell for it or whatev.

Cope, martial arts is peak

Diggy0:
Martial arts is slop.
Capoeira, CQC, Dragon Kung Fu, whatever ninja has.
The only decent one is sleeping carp and even then it's kinda mid, something something half assed port and melee combat being awful anyways because sec just has melee and ranged weapons that take away chunks of your hp in seconds.

Disregarding my opinion with "cope" when I say we shouldnt give a midround antag another martial art when people are malding like hell about ninja even having something outside of katana and shuriken for offensive.
Dragon should feel unique and fun, an actual threat rather than "le glorified mob that can spawn 2 ai mobs before it gets curbstomped by sec"

Someone got killed by dragon n immediately went to make a mald pr, because I really don't see why we should revert dragon back to making it weak / wizden weak, making it require 40 pop (which, maybe 2 or more of those persons are in sec) while also keeping security as strong as is right now (Waiter waiter more yamlslop weapons for sec that can end antags in 2 shots).

Anui-ElXx:
Not that deep bro.

I just think a DRAGON knowing Kung Fu would be rad as hell

NotActuallyMarty:
just adding this cause i want to have a think on how to merge this without having the entirety of upstream fucking up on god.

3PonPon3:
Personally, I'd say that yes, majority of times dragon fails is when sec fails to mobilize and cooperate. Although, it does feel more like a major antag than a minor antag, for how we have to arm and sometimes even arm crew.
I've had dragon rounds going well but I could still get beat because sec actually mobilized, and mobilized crew too once it was clear they need help due to low numbers

Personally, the only itchy part about current spess dragon, is the wing flap - it's an aoe stun "I win" button, that does not take any skill to use. Just press stun -> carp spawn = win.

Replace that stun with smthn else and imo, we're good

PR #826, opened 2026-04-09
Add PenetrationFactor to projectiles.

CliveOcelot:
## About the PR
Allows projectiles to modify how much Penetration Damage is taken when hitting an object, currently used by BlackIron slugs.

## Why / Balance
So bullets can penetrate further without needing to do ridiculous amounts of damage to a single target.

## Requirements

**Changelog**
:cl:
- tweak: BlackIron slugs now penetrate through even more walls.

### DIFF
--- a/Content.Server/Projectiles/ProjectileSystem.cs
+++ b/Content.Server/Projectiles/ProjectileSystem.cs
@@ -193,12 +193,12 @@ public sealed class ProjectileSystem : SharedProjectileSystem
             {
                 // Goobstation - Here penetration threshold count as "penetration health".
                 // If it's lower than damage than penetation damage entity cause it deletes projectile
-                if (component.PenetrationThreshold < penetratable.PenetrateDamage)
+                if (component.PenetrationThreshold < (penetratable.PenetrateDamage * component.PenetrationFactor)) // Omu, add penetration factor
                 {
                     component.ProjectileSpent = true;
                 }
 
-                component.PenetrationThreshold -= FixedPoint2.New(penetratable.PenetrateDamage);
+                component.PenetrationThreshold -= (FixedPoint2.New(penetratable.PenetrateDamage) * component.PenetrationFactor); // Omu, add penetration factor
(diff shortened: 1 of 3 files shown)


NotActuallyMarty (changes requested):
Give me a reason we need this because i don't understand why you cant have certain weapons with lower or higher penetration values as you need it. Unless its some highly specific bullshit?

Cygnye:
> Give me a reason we need this because i don't understand why you cant have certain weapons with lower or higher penetration values as you need it. Unless its some highly specific bullshit?

Marty I asked for this. It's for a specific use case and will be nice to have in general for stuff.

Thesia000 (approved):

NotActuallyMarty:
> > Give me a reason we need this because i don't understand why you cant have certain weapons with lower or higher penetration values as you need it. Unless its some highly specific bullshit?
> 
> Marty I asked for this. It's for a specific use case and will be nice to have in general for stuff.

What is the specific use case wherein you cant just add extra penetration to what you're shooting

NotActuallyMarty (changes requested):
Kay i figured this wrong i see this is now a multiplier to the 'damage' the bullet itself takes.

There's not really a point to this and makes keeping track of the damage / pen numbers harder, plus modifies upstream code, and just overall adds more stats and clutter

If you mention that this strategy is better than say, setting the pen health of  the bullet you are shooting to something ridiculous which in my head and testing has the same effect, i'd like it shown.

Images here are top par of reinforced wall (3 tiles) and bottom part (3 tiles also), with top part being your changes, and bottom part being penetrated using ridiculous pen health numbers instead.

Image

Image

NotActuallyMarty (approved):
Whatever nobody is willingly upstreaming besides me and clive atp and i was doing math in my head mid-argum ent and it is more comprehensible to give bullets "armor" and whatnot but listen my point at the time was that this was moot for this specific use case. Whatever.

PR #840, opened 2026-04-13
[FIX] Allow larger characters to block with shields properly.

CliveOcelot:
## About the PR
Removes the check for intersecting people when blocking (you push people out the way when you raise your shield anyway so this check isn't really needed), this makes it so that larger species can block using shields properly.

## Why / Balance
Larger species should be able to also block using shields.

## Technical details
Comments out a check for intersecting mobs

## Requirements

**Changelog**
:cl:
- fix: Larger characters should be able to block using shields again.

### DIFF
--- a/Content.Shared/Blocking/BlockingSystem.cs
+++ b/Content.Shared/Blocking/BlockingSystem.cs
@@ -191,7 +191,7 @@ public sealed partial class BlockingSystem : EntitySystem
             CantBlockError(user);
             return false;
         }
-
+        /* - Omu, this breaks stuff with larger characters, making them entierly unable to block.
         //Don't allow someone to block if someone else is on the same tile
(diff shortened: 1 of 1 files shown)


Content.Shared/Blocking/BlockingSystem.cs
```diff
@@ -208,6 +208,7 @@ public bool StartBlocking(EntityUid item, BlockingComponent component, EntityUid
             }
         }
 
+
```
CliveOcelot:
```suggestion
```

NotActuallyMarty (approved):
I don't like it either but worst comes to worst the boundingbox for checking characters just happens to be smaller than the character fixtures so its a shit fix for a shit problem that doesn't require me to think further on.

CerberusWolfie (approved):

PR #850, opened 2026-04-14
[FIX] Hopefully make the server updater work consistently.

CliveOcelot:
## About the PR
Basics of how this works, when a round ends (but has not gone to lobby yet, ex when the evac shuttle docks to CC), raise an event, then listen for this event, if the cvars for the updater are set and it is enabled, send a API request to Pterodactyl ro restart the relevant updater, this should hopefully restart in time and tell the game server to update before it would time out.

The event used is one I ported from EE for convenience.

## Why / Balance
At the moment we have a race condition with the updater.

## Technical details
See About the PR (or ask me)

## Requirements

**Changelog**
:cl:
- fix: Fixed the server failing to update.

### DIFF
--- a/Content.Omu.Common/CCVar/CCVars.Omu.cs
+++ b/Content.Omu.Common/CCVar/CCVars.Omu.cs
@@ -1,3 +1,4 @@
+using System.Runtime.InteropServices.Marshalling;
 using Robust.Shared.Configuration;
 
 namespace Content.Omu.Common.CCVar;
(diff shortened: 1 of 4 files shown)


Content.Omu.Common/CCVar/CCVars.Omu.cs
```diff
@@ -1,3 +1,4 @@
+using System.Runtime.InteropServices.Marshalling;
```
CliveOcelot:
```suggestion
```

PR #862, opened 2026-04-17
[PORT] Port TargetObjectiveImmuneComponent from DeltaV

CliveOcelot:
## About the PR
Ports TargetObjectiveImmuneComponent from DeltaV, makes two new variants of the mindshield implant (One also adds TargetObjectiveImmuneComponent, the other adds TargetObjectiveImmuneComponent  and AntagImmune), with the one that adds TargetObjectiveImmuneComponent  being given to BSO, NTR, ERT and other CC related roles.

## Why / Balance
People with a lifeline shouldn't be valid targets.

## Technical details
This contains a few things, including a system for implants to add components to whoever they are implanted into, two new filters which work in addition to edits to PickObjectiveTargetSystem to make people with TargetObjectiveImmuneComponent invalid targets.

Now are all of the filters needed alongside the edits to the system? I'm not sure. the it took a while of messing with them to get it to work, as filters seem entirely broken at the moment.

And before someone asks why `Content.Shared/_Omu/Mind/Filters/BodyMindBlacklistFilter.cs` is not in Omumod or the omu namespace, The filters system throws a decent few errors if it is not done the way it is. Why? Not a clue, but it does.

## Requirements

**Changelog**
:cl:
- tweak: BSO and NTR are no longer valid targets for Kill objectives (alongside all other CC roles, and nukies).


raincally:
does this apply to teach a lesson too? i understand prevent because You pretty much cant prevent them from getting to cc
i think you should still be able to feed the dignitaries a knuckle sandwich if the syndicate so desires

CliveOcelot:
> does this apply to teach a lesson too? i understand prevent because You pretty much cant prevent them from getting to cc i think you should still be able to feed the dignitaries a knuckle sandwich if the syndicate so desires

Since teach a lesson uses the same underlying shit to pick people, i'm pretty sure it prevents the "teach a lesson" objs as well (Granted i don't think you could complete those either since they would lifeline before they are fully dead, and you need someone to fully die once for teach a lesson to work)

raincally:
> > does this apply to teach a lesson too? i understand prevent because You pretty much cant prevent them from getting to cc i think you should still be able to feed the dignitaries a knuckle sandwich if the syndicate so desires
> 
> Since teach a lesson uses the same underlying shit to pick people, i'm pretty sure it prevents the "teach a lesson" objs as well (Granted i don't think you could complete those either since they would lifeline before they are fully dead, and you need someone to fully die once for teach a lesson to work)

i mean as long as the ntr doesnt get on the shuttle prevent from reaching cc is a greentext,, the objective doesnt make any sense tho
im pretty sure teach a lesson works fine but we should figure that out before deciding nothing bad can ever happen to dignitaries

CliveOcelot:
> > > does this apply to teach a lesson too? i understand prevent because You pretty much cant prevent them from getting to cc i think you should still be able to feed the dignitaries a knuckle sandwich if the syndicate so desires
> > 
> > 
> > Since teach a lesson uses the same underlying shit to pick people, i'm pretty sure it prevents the "teach a lesson" objs as well (Granted i don't think you could complete those either since they would lifeline before they are fully dead, and you need someone to fully die once for teach a lesson to work)
> 
> i mean as long as the ntr doesnt get on the shuttle prevent from reaching cc is a greentext,, the objective doesnt make any sense tho im pretty sure teach a lesson works fine but we should figure that out before deciding nothing bad can ever happen to dignitaries

It would be possible to do it on specific objs if filters were actually working consistently (they don't currently)

PurpleEI:
Do Heretic targets also use the same system? Because I was about to bounty something similar to this because its impossible to get the NTR or BSO to not lifeline out if you drag them anywhere near a rune.

ArtemisYK:
> Do Heretic targets also use the same system? Because I was about to bounty something similar to this because its impossible to get the NTR or BSO to not lifeline out if you drag them anywhere near a rune.

BSO and NTR can't use their lifeline if they're mansus grasped.

Content.Shared/_DV/Implants/AddComponentsImplant/AddComponentsImplantSystem.cs
```diff
@@ -0,0 +1,39 @@
+using Robust.Shared.Containers;
+using Content.Shared.Implants;
+
+namespace Content.Shared._DV.Implants.AddComponentsImplant;
+
+public sealed class AddComponentsImplantSystem : EntitySystem
+{
```
NotActuallyMarty: kill this we already have it as `ComponentsImplantSystem`

Content.Shared/_DV/Implants/AddComponentsImplant/AddComponentsImplantComponent.cs
```diff
@@ -0,0 +1,26 @@
...
+/// <summary>
+///     When added to an implanter will add the passed in components to the implanted entity.
+/// </summary>
+/// <remarks>
+///     Warning: Multiple implants with this component adding the same components will not properly remove components
+///     unless removed in the inverse order of their injection (Last in, first out).
+/// </remarks>
+[RegisterComponent]
+public sealed partial class AddComponentsImplantComponent : Component
+{
```
NotActuallyMarty: kill, see above.

Content.Server/Objectives/Systems/PickObjectiveTargetSystem.cs
```diff
@@ -62,6 +63,32 @@ private void OnSpecificPersonAssigned(Entity<PickSpecificPersonComponent> ent, r
...
+        {
+            args.Cancelled = true;
+            return;
+        }
+        if (HasComp<TargetObjectiveImmuneComponent>(targetMind.OwnedEntity.Value))
+        {
+            args.Cancelled = true;
+            return;
+        }
+        // END Omu
```
NotActuallyMarty: merge into one `if` and don't just assume the entity is the mind entity use `TryGetMind` at least, `HasComp` handles nullable var you don't need to pass by value.

Content.Shared/_Omu/Mind/Filters/BodyMindBlacklistFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+    public EntityWhitelist Blacklist = new();
+
+    protected override bool ShouldRemove(Entity<MindComponent> ent, EntityUid? exclude, IEntityManager entMan, SharedMindSystem mindSys)
+    {
+        if (ent.Comp.OwnedEntity is not {} mob)
+            return true;
+
+        var sys = entMan.System<EntityWhitelistSystem>();
+        return sys.IsBlacklistPassOrNull(Blacklist, mob);
+    }
```
NotActuallyMarty: we have `BodyMindFilter`? Its a whitelist instead of a blacklist which are the same thing if you use them. You can even check this method leading to the same thing. Use `!type: BodyMindFilter` with `inverted: true` instead if you want a real blacklist.

NotActuallyMarty (changes requested):
half of this PR is adding things that we already have, see if you can redo the changes you need with our current systems instead of duplicating things for no reason.

BramvanZijp:
@CliveOcelot Does this also apply to paradox clones? If not, that should be added IMO.

Thesia000 (approved):

NotActuallyMarty (approved):
i'm convinced this isn't necessary but honestly ill just check it myself later and revert if need be

BramvanZijp:
I will still ask, hows this work with paradox clones? Are those prevented from spawning as BSO/NTR? (or god forbid, admin characters, which I have seen happen)

PR #874, opened 2026-04-19
Lockable Morgues

Thesia000:
## About the PR
The Morge can now be locked by command members

## Why / Balance
Easier enfocments of DNRs

## Technical details
adds the lock and accessreaded comps to the morge and configs them, default is unlocked

## Media
image
Marty enjoy the image ^^

## Requirements

## Breaking changes
None

**Changelog**

:cl: Thesia
- add: Morgues can now be locked by command


raincally:
I like it but i think if we're gonna add an access reader you should be able to break them open with a structural weapon Cause right now theyre just indestructible im pretty sure

Thesia000:
good call raincall shall be adding that comp as well then

NotActuallyMarty (changes requested):
Should still be medical access. a visual lock indicator would be nice but i won't force it. often times DNR's can be unfounded, or medical needs something from the body, items can get locked inside, etc. Plus guarantee you folk are going to lock things for little reason leaving medical to scramble. Or in a brigmed setting a Hos will lock a prisoner that is expected to be revived, only for the doc needing to find them to unlock the morgue again and so on. I'm just expecting a lot of issues by making it default `command` lock.

It's fine as an addition, I would prefer this as default medical lock, then either a separate child prototype for different access, and or remapping to include separate morgues that are command / sec access for DNR bodies.

NotActuallyMarty:
also did not test make sure its not lockable/unlockable while open this is a very base proto

BramvanZijp:
I will note, if you default it to need med OR command access, and people want a specific slab to be used for a DNR, that can simply be handled via access config. 

Also, if we are making them breakable, we should add a way to build new morgue slabs too.

Thesia000:
self closing this till requested cahgnes are applied

PR #894, opened 2026-04-25
[FIX] fix uplinks being able to buy species restricted items.

CliveOcelot:
## About the PR
If BuyerSpeciesCondition fails to find a mind of appearance, it just returns false now, so that the species condition works. To note, this means that Pen uplinks, uplink implants, radio uplinks, and surplus crates cannot buy any of these items, as none of these set a Account.Owner, so the conditions cannot get the buyer.

## Why / Balance
Creates balance issues when people can get items that are balanced around a specific species using them, see beasts blood gourd. 

## Technical details
Change true to false

## Media
image
Left is pda uplink, right is uplink implant

## Requirements

## Breaking changes
Uplink implants, Pen uplinks (I think), surplus crates, and radio uplinks will no longer be able to buy species specific gear.

**Changelog**
:cl:
- fix: Fixed being able to bypass species requirements on uplinks

### DIFF
--- a/Content.Server/Store/Conditions/BuyerSpeciesCondition.cs
+++ b/Content.Server/Store/Conditions/BuyerSpeciesCondition.cs
@@ -42,10 +42,10 @@ public sealed partial class BuyerSpeciesCondition : ListingCondition
         var ent = args.EntityManager;
 
         if (!ent.TryGetComponent<MindComponent>(args.Buyer, out var mind))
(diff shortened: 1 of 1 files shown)


CliveOcelot:
after looking at uplinksystem this should only affect purchased uplinks (ex, uplink implant bought from pda uplink), or the radio uplink nukies spawn with, as neither uses SetUplink when spawned.

NotActuallyMarty (approved):

PR #896, opened 2026-04-25
Replace transits with Fugitive

CliveOcelot:
## About the PR
Disables transit prisoner entirely, ports the fugitive midround from DV

## Why / Balance
Admin vote regarding transits was to replace them with convicts from DV.

## Technical details
Disables transits showing in the menu, removes all of their job slots, use migration.yml to make all of their spawners be replaced with null. Then ports convict from DV.

Convicts will use the MidRoundAntagSpawner first when that's mapped, but until then it uses its fallback of a random vent.

## Media
The following is the sets they get to pick one of:
image
image

Example of the crimes paper that gets spawned in the convict's bag along with at every comms console:
image

## Requirements

**Changelog**
:cl:
- add: Ported the convict midround from DV
- remove: Removed transit prisoners


NotActuallyMarty (changes requested):
yah this still needs a lot of discussion but if you're removing content can you kill all the transit shit properly instead of leaving it in the codebase for no reason

CliveOcelot:
> yah this still needs a lot of discussion but if you're removing content can you kill all the transit shit properly instead of leaving it in the codebase for no reason

Removed it all

PR #913, opened 2026-04-30
[PORT] Port DV's epsilon tweaks

CliveOcelot:
## About the PR
Ports the update DV did to epsilon, now it is no longer "everyone's contracts are terminated", instead it overrides chain of command entirely to be central command personnel being the only commanding staff. It also adds a gamerule which disables power, and then when power comes back up, sets the alert to epsilon (And sounds cool while doing so)

## Why / Balance
Allows us to have an alert level for when we find command/sec to be stupid, and wish to step in, but do not wish for the entire crew to die (although it can still be used for that if needed).

## Requirements

**Changelog**
:cl:
- add: Ported DV's tweaks to the epsilon alert level.


CliveOcelot:
i hate mispressing, gimme a min to update the name

ArtemisYK:
I like this change.

Marikiii:
same here! maybe nerfing deathsquad a bit  or maybe even changing their color scheme would be cool too, later down the line that is

ArtemisYK:
> same here! maybe nerfing deathsquad a bit or maybe even changing their color scheme would be cool too, later down the line that is

DS themselves do not need changes, They're as OP as they are on EVERY server because they are MEANT to be that way. Funnily enough they are still VERY killable through a little creativity.
As for the colour scheme change, No.

Marikiii:
> > same here! maybe nerfing deathsquad a bit or maybe even changing their color scheme would be cool too, later down the line that is
> 
> DS themselves do not need changes, They're as OP as they are on EVERY server because they are MEANT to be that way. Funnily enough they are still VERY killable through a little creativity. As for the colour scheme change, No.

we need bright pink deathsquad you dont get it :3

Thesia000 (approved):

NotActuallyMarty (changes requested):
fix namespace

Content.Server/_DV/StationEvents/Components/EpsilonEventRuleComponent.cs
```diff
@@ -0,0 +1,28 @@
+using Content.Server.Power.Components;
+using Content.Server.StationEvents.Events;
+using Robust.Shared.Audio;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server.StationEvents.Components;
```
NotActuallyMarty:
```suggestion
using Content.Server._DV.StationEvents.Gamerules;
using Content.Server.Power.Components;
using Robust.Shared.Audio;
using Robust.Shared.Prototypes;

namespace Content.Server._DV.StationEvents.Components;
```

Content.Server/_DV/StationEvents/GameRules/EpsilonEventRule.cs
```diff
@@ -0,0 +1,73 @@
+using Content.Server.Power.Components;
+using Content.Server.Power.EntitySystems;
+using Content.Server.StationEvents.Components;
+using Content.Shared.GameTicking.Components;
+using Content.Shared.Station.Components;
+using Content.Server.AlertLevel;
+using Content.Server.Audio;
+using Robust.Shared.Audio.Systems;
+
+namespace Content.Server.StationEvents.Events
```
NotActuallyMarty:
```suggestion
using Content.Server._DV.StationEvents.Components;
using Content.Server.Power.Components;
using Content.Server.Power.EntitySystems;
using Content.Shared.GameTicking.Components;
using Content.Shared.Station.Components;
using Content.Server.AlertLevel;
using Content.Server.Audio;
using Content.Server.StationEvents.Events;
using Robust.Shared.Audio.Systems;

namespace Content.Server._DV.StationEvents.Gamerules
```

NotActuallyMarty (approved):
whatever

PR #930, opened 2026-05-05
Rebalances luxurium and luxury medipens

archee1:
## About the PR
Overhauls luxury medipens and the luxurium reagent to make more sense to players and be a more 'sensible' implementation of a combat medipen for lavaland.
Nerfs the survival medipens to only be effective on lavaland (which currently means they will do nothing)
This does not re-add the luxury medipen to the salvage vendor as they shouldn't be bought due to bugs (see the breaking changes section), Also disables the survival medipen from the salvage vendor

## Why / Balance
lux pens even with their current poisoning from luxurium essentially function as a combat medipen, their 20u omnizine as a baseline outheals all of the luxuriums side effects, they have so much tranex in them that they overdose you for the first 10s despite that unless you're a resomi being poisoned by the 20u dexalin/saline it gives you, you will outheal said tranex overdose.
Survival pens balanced for parity with luxurium.

The new damages luxurium heals puts the chemical in a better place for balance now only being useful for the typical situations found on lavaland.
Along with some needed overdose effects that now aren't stemming from the previous 40u overdose of 7 different medbay chems.
This change also means that resomi/avali are also no longer easily killed by one of these pens.

## Technical details
all yaml changes, 1 ftl change.

## Media
media outdated, see `_Lavaland/Reagents/medicine.yml`
image

## Requirements

## Breaking changes
the reagent condition `PressureThreshold` doesnt function, until that is fixed these new versions of the chems will have no effect in all situations.

**Changelog**
:cl:
- tweak: Luxury medipens no longer have 20u of 8 different chemicals.
- tweak: Luxury medipens (though disabled) cost twice as many salvage points to purchase.
- tweak: Luxurium is now a realistic medicine with its effects being only active on Lavaland.
- remove: Survival Medipens are no longer buyable at the salvage vendor, and also have no effect outside of lavaland.


CliveOcelot (changes requested):
If your re-doing the pen, remove all chems from it that are not luxurium, and balance luxurium accordingly (*ZERO* function when not on lavaland*, no damage, it just outright does not work.)

PR #953, opened 2026-05-10
Remapping of abductors shuttles

archee1:
## About the PR
Remaps the abductors shuttles interior by re-arranging its layout and placements of objects, all items previously there are still there, there is no change to the abductors kit or available items.
Better uses engineering to make the shuttle less gamey.
Lowers their shuttle's thrusters from 4 all directions + 5th forward with 3 gyros -> 3 all directions and 1 gyro.
Changes their power generation from 6 generators (30kw) to 2 rtgs + 1 shuttle APU (26kw)

## Why / Balance
Every time i looked at the original shuttle i would pull my hair out, it actively had 2 n2 miners and an o2 miner inside its walls connected to passive vents to fill the main hull, all of its electrical components were buried inside walls and in excessive amounts, 3 gyros for a shuttle with 0 directional care, 6 5kw generators instead of a 3 RTGs etc.
This gives abductors a more 'livery' style space to work from without changing any core balance behind their gameplay.

As for giving them a regular air canister instead of their o2 miner setup, there was no reason to devote as much space as it had to an atmos generator, a single canister of air is enough to keep their shuttle operable for in excess of 6 hours, caveat being if their hull is breached, but in those cases its likely gameover, abductors have very little EVA capabilties, if salv are on them their shuttle being depressurized for their abductions is the last thing on their minds.

## Technical details
2 map file changes

## Media
windows are reinforced uranium* image outdated.
image

## Requirements

**Changelog**


Dishwasher0 (changes requested):
Add at least one window so that crew can get into it without needing to demolish walls

Dishwasher0 (approved):

Willzile1 (approved):
Seems fine with me.
Handling is good, layout is easy to navigate.
All equipment seems to be the same.
only minor gripe is, plasma instead of uranium glass to stay on theme, but not a necessary change.

archee1:
> Seems fine with me. Handling is good, layout is easy to navigate. All equipment seems to be the same. only minor gripe is, plasma instead of uranium glass to stay on theme, but not a necessary change.

colours may have looked off in the image, they should be uranium

PR #991, opened 2026-05-17
[FIX] Fix people being able to print medipens for zero cost.

CliveOcelot:
## About the PR
Prevent anything with hypospray component from being inserted into a medical assembler as it causes it to bug out.

## Why / Balance
No more infinite medipens.

## Technical details
Adds a entity whitelist (blacklist) to microwave component, if it has anything in it then check if the item being inserted should be blacklisted, if it should be, display to the user that it doesn't fit, and don't insert it. (if blacklist is null then allow insertion as usual)

## Requirements

**Changelog**
:cl:
- fix: Fixed being able to make infinite medipens for free.

### DIFF
--- a/Content.Server/Kitchen/Components/MicrowaveComponent.cs
+++ b/Content.Server/Kitchen/Components/MicrowaveComponent.cs
@@ -69,6 +69,7 @@ using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom; // Fro
 using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype;
 using Content.Shared.Kitchen.Components; // Frontier
 using Content.Shared.Kitchen; // Frontier
+using Content.Shared.Whitelist; // Omu
 
 namespace Content.Server.Kitchen.Components
 {
@@ -213,6 +214,14 @@ namespace Content.Server.Kitchen.Components
         [DataField]
         public MicrowaveUiKey Key = MicrowaveUiKey.Key;
         // End Frontier
+
+        // Begin Omu
+        /// <summary>
+        /// A blacklist of components/tags that should not be allowed to be inserted into a microwave.
+        /// </summary>
+        [DataField, ViewVariables(VVAccess.ReadWrite)]
+        public EntityWhitelist MicrowaveBlacklist = new();
+        // End Omu
     }
 
     public sealed class BeingMicrowavedEvent : HandledEntityEventArgs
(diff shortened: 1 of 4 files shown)


Aureu-K (approved):

NotActuallyMarty (changes requested):
Replace `isblacklistpass` with `iswhitelistpass`, `isblacklistpass` is removed upstream and is the same method. add a reminder for yourself to fix this later hypospray component is entirely removed upstream too replaced with injector. No need to fix that now but yeah, remember for later.

also id prefer to kill `components` and use componentregistration field instead but thats larp and on whitelistsystem tbf,  if you're too lazy keep as is.

selfmerge once addressed.

Content.Server/Kitchen/Components/MicrowaveComponent.cs
```diff
@@ -213,6 +214,14 @@ public sealed partial class MicrowaveComponent : Component
         [DataField]
         public MicrowaveUiKey Key = MicrowaveUiKey.Key;
         // End Frontier
+
+        // Begin Omu
+        /// <summary>
+        /// A blacklist of components/tags that should not be allowed to be inserted into a microwave.
+        /// </summary>
+        [DataField, ViewVariables(VVAccess.ReadWrite)]
```
NotActuallyMarty: also on god you do not need to add vvaccess anymore to things.
```suggestion
        [DataField]
```

Content.Server/Kitchen/EntitySystems/MicrowaveSystem.cs
```diff
@@ -444,6 +446,14 @@ private void OnInteractUsing(Entity<MicrowaveComponent> ent, ref InteractUsingEv
                 return;
             }
 
+            // Omu start
+            if (_whitelistSystem.IsBlacklistPass(ent.Comp.MicrowaveBlacklist, args.Used))
```
CliveOcelot:
```suggestion
            if (_whitelistSystem.IsWhitelistPass(ent.Comp.MicrowaveBlacklist, args.Used))
```

Resources/Prototypes/_NF/Entities/Structures/Machines/mirowave_recipe_assemblers.yml
```diff
@@ -228,6 +228,9 @@
     foodDoneSound: /Audio/Effects/Shuttle/radar_ping.ogg
     loopingSound: /Audio/_NF/Machines/medical-assembler.ogg
     key: MedicalAssemblerKey
+    microwaveBlacklist: # Omu start
+      components:
+      - Hypospray # Omu end
```
CliveOcelot:
```suggestion
      - Hypospray # NOTE: Will need fixing with upstream, Omu end
```

PR #1009, opened 2026-05-22
Fix: IPC Health Examine not showing Body Part health

RadsammyT:
## About the PR

Fixed the IPC Health Examine not showing body part status.

## Technical details
The cause was from an upstream removing an IPCs bones (https://github.com/Goob-Station/Goob-Station/pull/4905). Shitmed doesn't account for a species having no bones so it doesn't show anything. The fix was to make only a `BodyPart` component required for anything to show. If it is boneless, it defaults to Normal Bone Severity (which doesn't show anything).

My day has been ruined and my disappointment is immeasurable. I am going to beat something with a rock now.

## Media

image

## Requirements

**Changelog**

:cl:
- fix: The state of IPC body parts will now show up properly when examining an IPCs health.


NotActuallyMarty (changes requested):
coulda sworn this was due to slimes or something instead.

`Try`  in `TryComp` implies we try to find the comp. we can thus fail to find it and it be null. marking the bone as nullable and then showing normal boneseverity is not the right way to fix this I'd prefer  to see either none bones (no text relating to bones entirely) or just text that says theyve got none. Just add an extra enum or something.

kinda larp but if this has existed for this long we can fix it properly atp

NotActuallyMarty (approved):
yeah sure

PR #1028, opened 2026-05-25
[FIX] Make bingles actually take ion damage.

CliveOcelot:
## About the PR
When looking at bingles damage modifier i noticed they were meant to take 2x ion damage, however their damage container doesn't support ion damage. So this adds a new damage container that is the same as biological, but with electronic (ion) damage added to it.

## Why / Balance
If their damage modifier says they are meant to take 2x ion damage, they should be able to take ion damage.

## Technical details
Add a new damage container, swap bingles to use the new damage container.

## Media
bingle's damage modifier:
image

## Requirements

**Changelog**
:cl:
- fix: Fixed bingles being unable to take ion damage.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Mobs/Player/bingle.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Mobs/Player/bingle.yml
@@ -108,7 +108,7 @@
     overtime: 15
     sound: /Audio/Weapons/egloves.ogg
   - type: Damageable
-    damageContainer: Biological
+    damageContainer: BiologicalElectrical # Omu, was Biological
     damageModifierSet: Bingle
   - type: FootstepModifier
     footstepSoundCollection:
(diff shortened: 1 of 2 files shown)


Thesia000 (approved):

MarkerWicker:
This appears to be a problem upstream as well. Why not send this merge request to Goobstation?

Dishwasher0:
> This appears to be a problem upstream as well. Why not send this merge request to Goobstation?

We had whole conversation about sending people to goob git hub 
We don't force anyone to do anything 
And also goob can just down stream it if they want

Sarahon:
> > This appears to be a problem upstream as well. Why not send this merge request to Goobstation?
> 
> We had whole conversation about sending people to goob git hub We don't force anyone to do anything And also goob can just down stream it if they want

You act as if this is like the end of the world tbh. Upstreaming 'fixes' is beneficial for Omu cause Goob is Omu's upstream. People being stubborn about it is kinda just on them. The more weird things they do the more annoying upstreaming will become due to avoiding upstream. 

People have no reason to avoid upstreaming to wizden, we just know the outcome to that.

Aureu-K (approved):

PR #1049, opened 2026-05-31
Add `on` and `this` Toolshed Commands

ishkab:
## About the PR
This PR adds two Toolshed commands (8 total subcommands) for filtering entities based on map or grid predicates. All six subcommands of `on` are invertible with the usual `not` construction. These are mainly convenience functions for administration.

- `on:grid <gridid>` filters an IEnumerable<EntityUid> to only those entities that are on the grid with the ID given.
- `on:thisgrid` filters an IEnumerable<EntityUid> to only those entities that are on the same grid as the invoker.
- `on:gridof <ent>` filters an IEnumerable<EntityUid> to only those entities that are on the same grid as the entity given.
- `on:map <mapid>` filters an IEnumerable<EntityUid> to only those entities that are on the map with the ID given.
- `on:thismap` filters an IEnumerable<EntityUid> to only those entities that are on the same map as the invoker.
- `on:mapof <ent>` filters an IEnumerable<EntityUid> to only those entities that are on the same map as the entity given.
- `this:grid` returns the grid that the invoker is on.
- `this:map` returns the ID of the map that the invoker is on.

## Why / Balance
- These are Toolshed filters that I personally want.
- They may be generally useful to anyone who uses Toolshed, e.g. in execs. With `on:thisgrid`, one can easily target only entities on-station, for instance.
- Mappers should find the `on` command useful for removing stray grids created in error during the mapping process, as in `entities on:thismap with MapGrid not with BecomesStation delete`.

## Technical details
I originally implemented this way out of code style with the rest of the repo. I think the code is less readable now, but it conforms.

I added OnCommand and ThisCommand along with three Toolshed errors. The commands require DEBUG permission.

## Requirements

## Breaking changes
I tested this very briefly and found it to be in enough of a working order to PR. If bugs are discovered, they should be strictly contained to the execution of the commands - a situation which makes for easy avoidance until a patch can be issued.


feistyfedora:
Stale

harryhulse12-alt:
Fish if you can show me how this works boss

PR #1051, opened 2026-06-01
Insulated items are tagged on examine

RedstonewolfX:
Items will tell you if they provide protection from electric shocks.

## About the PR

Shift clicking an item that protects you from electric shocks will show a text popup explaining the feature, like how sunglasses tell you they protect from flashes, or over-ears from flashbangs.
Budget Insulated Gloves claim to be protective regardless of whether they actually work or not.
Conductive and Explosive Gloves will lie and display as insulative, making them indistinguishable from Insulated Gloves from the outside.

## Why / Balance

This is simply a quality of life feature. Some items say they're insulated when they aren't, or are insulative but don't make it obvious.

## Technical details

Adds an event listener to `ElectrocutionSystem` that I pulled from the code for flash immunity. When an item with `InsulatedComponent` is Examined, it shows that it "Protects from most electrical shocks."
(I tried to make items with random insulation instead say that they were unreliable, but couldn't figure it out)
Due to how `InsulatedComponent` is implemented, items that deal more shock damage like Conductive Gloves will still look like insuls.
Chameleon gloves do not show, but this is also a problem with chameleon Sun Glasses, so I didn't bother.

## Media

image

https://github.com/user-attachments/assets/8d828d8f-613a-4867-b617-3235d94d64b6

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: You can now see if something protects you from electrocution by examining it


RedstonewolfX:
I forgot to check if this works with the Insulated enchantment gulp

archee1:
upstream has a thing like this that we dont have yet, may want to look at that and port it instead, it uses the same system that seeing armor values uses to show its insulated with a lightning bolt

Dishwasher0:
> upstream has a thing like this that we dont have yet, may want to look at that and port it instead, it uses the same system that seeing armor values uses to show its insulated with a lightning bolt

we will get up up stream (wizden one) when goob's done and we do our upstream so soon TM

RedstonewolfX:
Oh, that’s neat, and better than this system

PR #1080, opened 2026-06-10
Unrevivable comp changes

Thesia000:
## About the PR
Adds a new datafield to the unreivable component and adapts the CPR system

## Why / Balance
Makes unrevivable work in line with how the heath analyser descibes it to weak for the defib

## Technical details
Adds a new datafield to the unreivable component and adapts the CPR system

## Requirements

## Breaking changes
- Adds a new datafiled to the Unrevivable component
- Touces the CPR system

**Changelog**

:cl:
- tweak: CPR can revive unreviable trait again.

### DIFF
--- a/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
+++ b/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
@@ -134,9 +134,8 @@ public sealed class CPRSystem : EntitySystem
             && TryComp<MobStateComponent>(args.Target, out var state)
             && _mobThreshold.CheckVitalDamage(args.Target.Value, damageableComponent) < threshold)// GoobStation
         {//OMU start
-            if (HasComp<UnrevivableComponent>(args.Target))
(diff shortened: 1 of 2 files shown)


ishkab (changes requested):
Instead of calling HasComp and then TryComp, use TryComp in place of HasComp: its return value should tell you whether the entity has the component.

```cs
if(TryComp<UnrevivableComponent>(args.Target, out var unrevComp))
{
  if(!unrevComp.CPRBlock) // if we get here, we know unrevComp is not null - we shouldn't need the null-forgiving operator
  {
    _mobStateSystem.ChangeMobState(args.Target.Value, MobState.Critical, state, performer);
  }
}
else
{
  _mobStateSystem.ChangeMobState(args.Target.Value, MobState.Critical, state, performer);
}
```

ishkab (approved):
Seems good to me. I haven't tested this locally but my estimation is that it should work as it reads.

Aureu-K (approved):

PR #1084, opened 2026-06-11
Free  pun pun

raincally:
## About the PR

Lowers pun pun role timer to 5 hours overall instead of 15 bartender
## Why / Balance

I need more  monkey business
Shitters and raiders are not a reason to increase role requirements that much,, its an admin issue let admins handle it
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Pun pun role timer requirements lowered. 15 bartender hours -> 5 overall hours

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/NPCs/pets.yml
+++ b/Resources/Prototypes/Entities/Mobs/NPCs/pets.yml
@@ -1170,7 +1170,7 @@
     requirements: # Goobstation - ghost roles requirements
     - !type:RoleTimeRequirement
       role: JobBartender
-      time: 54000 #15 hour
+      time: 3600 # Omu, 1 hour
     mindRoles:
     - MindRoleGhostRoleNeutral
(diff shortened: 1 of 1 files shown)


Thesia000 (changes requested):
This needs be be brought up in a staff discussion first like any timechanges.

jckspjf:
Even pre-staff discussion, shouldn't this just be the same requirements as bartender? If not only a tad bit higher instead of 5 hours?

Anui-ElXx:
15 hours of bartender is crazy. Almost nobody has 15 hours of bartender, and the few who has prefer to play as bartender instead (obviously).

That basically means that we never get go see Pun-Pun alive. Free my boy.

madcaker:
Something like 5 bartender and 15 overall might make sense for this? Depends on if people think the 15 overall would be enough to reduce raiders.

feistyfedora (approved):
Looks good to me

Thesia000 (approved):

PR #1102, opened 2026-06-17
No more SaliOxan Mix

Willzile1:
## About the PR

Mixing SalicylicAcid and Oxandrolone will decompose into some of it's base components, plus space cleaner.
## Why / Balance

Lazy doctors and chemists will mix Sali and Oxan to have an omni injection that can be used on anyone for anything. Also stops people becoming nigh unkillable with that same mix.
## Technical details

Added SalicOxanDecomp reaction in single_reagent.
## Media

Result of mixing 10u of SalicylicAcid and Oxandrolone
image

Approximate damage values
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: SalicylicAcid and Oxandrolone can no longer be mixed together.


archee1:
personally i think this is the wrong approach, the core issue with these chems even in isolation is they are far too strong in combat situations, their specialty of 'being bad for low amounts' is better suited as 'being good for high amounts' in examples of explosives spacing or fires.

This also doesnt address that you can mix up something such as Salic/Punct and now have a 4/4/8 healing rate chem mix, i get that the idea is to adress 'lazy mixes' but the same is possible by swapping the oxandrolone out for dermaline, which has relatively as decent potency to oxandrolone.

Sure these chems are in the 'high end' of production though to chemists worth their salt this doesnt mean they take any longer than 4 minutes to make, and doesnt validate their power.

I think a better solution would be something similar to what wizden did to their tier 2 brutes a good while back, halving the metabolism rates and healing so that it takes twice as long to get the same effect, reducing its effect for active combat. perhaps raising the thresholds from 25 -> 50 also, so that its further dampened as a low end healer for someone that has had bone damage and would then go to rely on topicals.

Willzile1:
Less about their combat use, I don't mind that aspect, what I am trying to combat is a doctor just injecting a patient with 7.5 of both chems and instantly healing any and all damage without even checking them. I want doctors to need to actually check what damage their patients have instead of shotgunning them with no consequence.

Sali and Oxan have literally zero downsides to their use. Tier ones have the low healing potential, tier twos have the specialized brutes that are unable to be mixed, tier fours have the comically low OD amount, but tier threes have nothing. Tier threes are the best of all of them by far, they have the highest healing potential, they work incredibly fast, and have a high OD amount. Combining the two just make it nigh impossible to kill someone with any non-exotic damage types.

I still want to reward players for making it, but I don't want it to keep being the panacea it currently is. I want it to keep it's combat capabilities, but remove the triviality of it's use. Yes, you can swap out the Oxan for Darma, but that comes with a reasonable OD and less healing potential.

I want to break this one specific meta mix that makes people very hard to kill.

PR #1116, opened 2026-06-24
[Bounty] Bingle tweaks

derg4:
## About the PR

* Bingle pit now spawns bingles directly when previously you had to request the bingle via ghost roles
* Changed the ghost role raffle time to become a bingle to be 10 seconds
* Bingle pit no longer takes structural/caustic damage (changed to Inorganic instead of StructuralInorganic)

## Why / Balance

The discussion I've been a part of is [here in the relevant bounty thread in the Omu discord](https://discord.com/channels/1406768242843979816/1517044154876563497) but it wouldn't surprise me if there has been more.

For the pit spawning bingles directly and ghost role raffle time change, the idea is to make it easier and quicker to enter the game as a bingle. A few people in the bounty thread expressed the feeling that the default 25 second timer is too long. Also, the short raffle timer is intended for spammed roles like carp as noted [here](https://github.com/ProjectOmu/OmuStation/blob/master/Resources/Prototypes/GhostRoleRaffles/settings.yml#L43).

There was also some frustration about how easy it is to destroy a bingle pit with a single explosive. This is why the damage container was changed. This part is a buff to bingles, as it doesn't take structural or caustic damage anymore and so would be harder to deal with. In my test, a breaching charge now does just the 50 of each brute, so 150/600 health that it has. This shouldn't affect the other ways people have of dealing with it, though.

## Technical details

Changed the damage container in Binglepit.yml from `StructuralInorganic` to `Inorganic`
Changed the raffle settings in bingle.yml from `default` to `short`

## Media

https://www.youtube.com/watch?v=wdTY4araQlw

## Requirements

## Breaking changes

**Changelog**

:cl: Derg
- tweak: Bingle ghost roles no longer need to be requested and raffle time shortened.
- tweak: Bingle pit no longer takes structural damage.

### DIFF
--- a/Content.Goobstation.Common/Bingle/BinglePitComponent.cs
+++ b/Content.Goobstation.Common/Bingle/BinglePitComponent.cs
@@ -68,7 +68,7 @@ public sealed partial class BinglePitComponent : Component
     public SoundSpecifier FallingSound = new SoundPathSpecifier("/Audio/Effects/falling.ogg");
 
     [DataField]
-    public EntProtoId GhostRoleToSpawn = "SpawnPointGhostBingle";
+    public EntProtoId GhostRoleToSpawn = "MobBingleRandom"; // Omu
 
     /// <summary>
     /// how many bingles to spawn on pit spawn
--- a/Resources/Prototypes/_Goobstation/Entities/Mobs/Player/bingle.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Mobs/Player/bingle.yml
@@ -53,7 +53,7 @@
     mindRoles:
     - MindRoleGhostRoleTeamAntagonist
     raffle:
-      settings: default
+      settings: short
   - type: Sprite
     drawdepth: Mobs
     sprite: _Goobstation/Mobs/Bingle/bingle.rsi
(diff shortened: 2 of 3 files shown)


CliveOcelot:
If your doing this, it would be good to make it also rather than doing the whole you have to do request, then raffle to take control, just make it spawn the bingle, and let it use its AI till someone raffles for it

derg4:
You know, I do agree it would be better like that 100%, I can try to look into it

derg4:
Would it be better to change the bingle raffle to be even shorter or have no timer at all?

derg4:
I've updated the PR to have the bingle pit spawn the bingles directly, and updated* the test video in media to show all changes

derg4:
Actually, this might interfere with how the pit upgrades bingles (or does it? I don't actually understand how the upgraded bingles spawn in in the first place...).

EDIT: After growing the pit and examining a bingle (pre-my change), they appear the same but have the upgraded toggle on in their Bingle component. Is this a bug?

CliveOcelot (approved):

Thesia000 (approved):

PR #1131, opened 2026-06-27
Importent Xelthia bugfix

SirCannonFodder:
## About the PR

changed the base Xelthia mob to use Biological damage container rather then the Xelthia one
## Why / Balance

- So turns out Medical HUDs and tropicals use damage containers for checking if they can do there respective things. Not like you know... Tags? WHY doesn't just look for a tag on damage containers or the mob itself is beyond me.
- Xelthia are still fully immune to caustic damage because of how shitmed works with the body parts using the Xelthia damage container.
## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl:
- fix: Fixed a bug that stopped Xelthia from using topicals and being read by medical HUDs

PR #1134, opened 2026-06-27
Xelthia fix caustic damage attempt #2

SirCannonFodder:
## About the PR

reverts PR https://github.com/ProjectOmu/OmuStation/pull/1131
sets xelthia base mob do use biological damage container since it's required for medHUD and topicals to work.
## Why / Balance
working medicalHUD and topicals AND xelthia total caustic immunity.
A bit janky of a solution since it relies on how shitmed stores damage in body parts rather then the mob's damage container. FeistyFedora asks I could make a new system that could be applied to other cases in the future for this rather then this fix but C# right now that is beyond me.

- Why damage container rather then a damage modifier? because chemicals are the primary source of caustic damage and they bypass damage modifiers. I will the check the metabolizer code again to see if there's just a flag I can put on the modifier to have it apply to chems but I don't think there is.
- why didn't medicalHUDs and topicals work before? turns out there's a hardcoded whitelist that topicals and HUDs look at when checking if they can interact with a damage container.
- How does it work? The base mob uses the normal biological damage container and the body parts use the xelthia damage container. HUDs and topicals see the see the biological damage container and work as they should. Since shitmed stores damage inside the body part's container rather then mob's it allows xelthia to enjoy full caustic immunity while preserving HUD and topical functionality.
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed xelthia not being able to be read by medHUDs and use topicals while maintaining there caustic immunity


pinktanuki:
There is nothing noted in your change log

feistyfedora:
Fixed your title since "Bugfix 2" says nothing about the PR. Please be more descriptive.

Thesia000 (changes requested):
leave the damage modifier bit in, as a fallback

Thesia000 (approved):

CliveOcelot (approved):

PR #1170, opened 2026-07-03
Some QoL for PTLs

archee1:
## About the PR
removes the 'sale price' of the PTL being adjusted for its stored power to prevent people getting confused by price analysing it when they're meant to be multitooling it to withdraw money, it now always is analzyed as 1500$.

makes PTLs 5x heavier, decreases their hitbox by .3 metres in all directions, moves it to the machine layer instead of wall layer (less annoying to interact with if things get partially placed inside such is the case with watertanks/watercoolers if you're familiar)
increases the range of the laser fired and adds structural damage to prevent them being fired indoors. (see media for hitbox)

## Why / Balance
The changes to the laser projectile should prevent any indoor use, though i can see reason to skip on the added range if theres worries of a slightly longer laser projectile constituting towards bad server performance, though consider worst case is 10 PTLs firing their full length its 500 tiles of raycasting spread out over 10 seconds.

## Technical details
yaml slop

## Media
image

## Requirements

**Changelog**
:cl:
- tweak: PTLs are slightly smaller than they were before and weigh a more sensible amount for their size with their fired laser being more destructive.
- tweak: PTLs no longer gain price as their battery fills up and instead have a static price.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Structures/Machines/power_laser.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Structures/Machines/power_laser.yml
@@ -17,8 +17,8 @@
     mode: SnapgridCenter
   components:
   - type: PTL
-    minEnergyPerShot: 5e5 # Omu
-    maxEnergyPerShot: 3e6 # Omu 3MJ (300kw/s), 404spesos 10/s
+    minEnergyPerShot: 25e4 # Omu
+    maxEnergyPerShot: 3.5e6 # Omu 3.5MJ (@300kw/s, 404spesos 10/s), this is throttled to 3e6 by charge rate intentionally.
     active: false
     baseBeamDamage:
       types:
@@ -44,7 +44,7 @@
   - type: RadiationSource
   - type: Appearance
   - type: PowerNetworkBattery
-    maxChargeRate: 500000 # Omu, 500kW
+    maxChargeRate: 300000 # Omu, 300kW
   - type: Battery
     startingCharge: 0
     maxCharge: 25000000 # Omu, 25MJ (50s of 500kW)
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/_Goobstation/Entities/Structures/Machines/power_laser.yml
```diff
@@ -17,10 +17,13 @@
     mode: SnapgridCenter
   components:
   - type: PTL
+    minEnergyPerShot: 25e4 # Omu
```
feistyfedora: This isn't a real field 
Causing test fails
I think you meant MinShootPower?

feistyfedora (changes requested):

Resources/Prototypes/_Goobstation/Entities/Structures/Machines/power_laser.yml
archee1: most likely yeah myb

feistyfedora (changes requested):
I understand these are related to each other, but this should be two separate PRs. One for the SMES changes and another for the PTL changes.

archee1:
split the SMES change up into [PR#1178](https://github.com/ProjectOmu/OmuStation/pull/1178)

feistyfedora (changes requested):
PTL should have to be monitored and should cause power issues if not set up correctly. 
It is a huge money laser that needs to be carefully setup. 
Please remove the power related updates. 

The other QoL changes are okay thought.

feistyfedora (approved):
Looks good to me

CliveOcelot (approved):
looks good to me

PR #1243, opened 2026-07-17
Unmute Blinking For Mute People

codeduck101:
## About the PR

blinking now makes a sound if you're muted
added the option to unmute other vocal/general sounds with a new datafield

## Why / Balance

done as a bounty

it makes no sense why blinking should be muted, as it's a body sound
i went with the whole datafield solution for futureproofing if we add in other sounds that shouldn't be muted

## Technical details

added data field `OverrideMute` to `Content.Shared/Chat/Prototypes/EmotePrototype.cs`
changed the check in `Content.Server/Speech/Muting/MutingSystem.cs` for muting sounds to additionally check that mute shouldn't be overridden

## Media

https://github.com/user-attachments/assets/9ebf7668-7235-4f61-b4c0-4c0c1ee48804

## Requirements

## Breaking changes

**Changelog**

:cl: Duuckie
- add: Mimes and other mute characters can make noise by blinking.

### DIFF
--- a/Content.Server/Speech/Muting/MutingSystem.cs
+++ b/Content.Server/Speech/Muting/MutingSystem.cs
@@ -38,7 +38,7 @@ namespace Content.Server.Speech.Muting
                 return;
 
             //still leaves the text so it looks like they are pantomiming a laugh
-            if (args.Emote.Category.HasFlag(EmoteCategory.Vocal))
+            if (args.Emote.Category.HasFlag(EmoteCategory.Vocal) && !args.Emote.OverrideMute) // Omu, adds manual overriding of mute
                 args.Handled = true;
         }
 
--- a/Content.Shared/Chat/Prototypes/EmotePrototype.cs
+++ b/Content.Shared/Chat/Prototypes/EmotePrototype.cs
@@ -43,6 +43,14 @@ public sealed partial class EmotePrototype : IPrototype
     [DataField]
     public bool Available = true;
 
+    // Omu edit, added for handling blinking
+    /// <summary>
+    ///     Allows for overriding of mute trait for vocal/general emotes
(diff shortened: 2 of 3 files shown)


PhosphateWithManyOs (changes requested):
you can do this with YAML just by changing the category to category: hands instead of being vocal. It doesn't exaccctly make sense but stuff like tail thump follows the same.

codeduck101:
> you can do this with YAML just by changing the category to category: hands instead of being vocal. It doesn't exaccctly make sense but stuff like tail thump follows the same.

[see thread](https://discord.com/channels/1406768242843979816/1525249312483967126/1527767826176147506) where we tried that solution and wound up finding that
1. this places blink in the hands category
2. if you lose both your hands you also lose the ability to blink

additionally, this is a future-proofed solution. we could even move tail thump to use this system to avoid the "no hands prevents the emote sounds from playing" issue

PhosphateWithManyOs (approved):
shit I guess

codeduck101:
> instead of being vocal

as a notice, blinking is in "General"
all "General" emotes are treated like "Vocal" for the purposes of muting because all "Vocal" emotes are set up as a flag inside of a byte and "General" emotes are treated as having every flag; therefore, when the mute system checks for "Vocal" emotes, it's checking for a flag that both "Vocal" and "General" emotes have

here's the relevant enum:
```
public enum EmoteCategory : byte
{
    Invalid = 0,
    Vocal = 1 << 0,
    Hands = 1 << 1,
    Farts = 1 << 2, // Goobstation - Fart Emotes
    General = byte.MaxValue
}
```

PhosphateWithManyOs:
yeah that was a typo on my end, general is "muted" because of death gasp.

ishkab (approved):
I personally like these changes.

PR #1244, opened 2026-07-17
Fix Photophobia shader crash on GLES/compat-mode renderers

Not-Solid:
## About the PR
Fixes the Photophobia shader (`Resources/Textures/_Omu/Shaders/photophobia.swsl`) crashing the client on startup under the engine's GLES/ANGLE compatibility-mode renderer. Changes the `2`/`1` int literals inside float expressions to `2.0`/`1.0` — 3 lines, nothing else.

## Why / Balance
int literals mixed with float operands (`pow(2, ...)`, `clamp(v, 0.5, 1)`) compile fine on desktop OpenGL but are a hard error under GLSL ES, which is what the engine's ANGLE/compat-mode renderer uses (auto-forced on Qualcomm/Snapdragon ARM devices due to their broken native OpenGL driver). Since shader prototypes are compiled eagerly on client startup, this crashed the client for every player on such hardware before they could even reach the main menu — regardless of whether they had the Photophobia trait.

## Technical details
Only `Resources/Textures/_Omu/Shaders/photophobia.swsl` is touched (3 lines, int literal → float literal). No YAML or C# changes needed — `Resources/Prototypes/_Omu/Shaders/shaders.yml` is the only reference to this shader and stays as-is.

Test plan (no ARM hardware needed — the strict GLES compiler can be forced on any machine):
- `dotnet run --project Content.Omu.Client -- --cvar display.compat=true`

## Media
Not required

## Requirements

## Breaking changes
None.

**Changelog**
:cl:
- fix: Fixed the client crashing on startup for players on GLES/compatibility-mode renderers (e.g. Qualcomm/Snapdragon ARM devices) due to a shader compile error in the Photophobia effect.

### DIFF
--- a/Resources/Textures/_Omu/Shaders/photophobia.swsl
+++ b/Resources/Textures/_Omu/Shaders/photophobia.swsl
@@ -12,9 +12,9 @@ void fragment() {
     // Those arbitrary -0.5, 0.5, and 0.10 values are my attempt at dealing with the fact that humans percieve different colours differently.
     // There's probably a better way to do that. But I don't know much of vector math, and I've never done math with matrices in my life. So you get this.
     // Plus, I think there's a beauty in only blowing out one colour channel instead of all three. It looks good.
-    color.r = pow(color.r, pow(2, 2*(0.5-clamp((lightcolor.r - 0.05)*effectStrength, 0.5, 1))));
-    color.g = pow(color.g, pow(2, 2*(0.5-clamp((lightcolor.g + 0.05)*effectStrength, 0.5, 1))));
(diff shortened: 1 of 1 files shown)


Not-Solid:
Integrationtests did not fail due to my PR. The cause seems to stem from unstable test cases.

feistyfedora:
> Integrationtests did not fail due to my PR. The cause seems to stem from unstable test cases.

Rerunning

ishkab (approved):
Looks incredible, honestly. Shouldn't be any problem here imo.

feistyfedora (approved):
Looks good to me

PR #1315, opened 2026-07-28
Fix CPR Logic & Missing Text

codeduck101:
## About the PR

fixed a logic error allowing you to start CPR when your patient has a mask on
fixed people being able to continue doing CPR even when they're not supposed to be able to start it
added appropriate text popups when you're unable to do CPR and for what reason

## Why / Balance

i JUST found out that you're not supposed to be able to perform CPR if you/the victim has a mask on. (this mean you're using mouth-to-mouth. yucky)
having proper text (and the proper mask interactions) makes things WAY less confusing for the average doctor

## Technical details

changes are to EE's `CPRSystem.cs` for handling mask logic and getting CPR to stop
changes to wizden `IngestionSystem.API.cs` to add additional overloads of `HasMouthAvailable` to provide nullable `out` for possible messages (the message is `null` if the message isn't set by being out of range or having a face blocker) and nullable `out` for the ingestion blocked (used for proper text reasons of what's actually preventing you from doing CPR)
added omu-specific `cpr-training-component.ftl` to avoid messing with anything on EE's version

for the mask logic fix, simply switched two ~~parameters~~ ARGUMENTS i meant arguments
for the text to show up, i added the overloads of `HasMouthAvailable`, changed the checks to use the new version with a nullable `out` for the message and blocker, and split the checks for the performer and target to have unique text for each explaining who exactly has the problem
for the CPR to stop when you're not supposed to do CPR, i abstracted the whole-ass check in `StartCPR()` to helper method  `ShouldCPRStop()` that would check the reasons why you shouldn't be able to start and show the text popup corresponding to that reason. after that, i added a check for `ShouldCPRStop()` in the callback `OnCPRDoAfter` which stops you from doing CPR at the end of the current CPR cycle

## Media

https://youtu.be/1G0f6wmot3o

## Requirements

## Breaking changes

**Changelog**

:cl: Duuckie
- tweak: CPR now has proper text popups for when you're not able to do it.
- fix: CPR now stops you from continuing if you're not meant to be able to start CPR. You also can no longer do CPR to people who have masks on.

### DIFF
--- a/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
+++ b/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
@@ -16,6 +16,7 @@ using Content.Shared.Atmos.Rotting;
 using Content.Shared.Damage;
 using Content.Shared.DoAfter;
 using Content.Shared.Inventory;
+using Content.Shared.Nutrition; // Omu
 
 using Content.Shared.Medical.CPR;
 using Content.Shared.Mobs;
@@ -44,6 +45,7 @@ public sealed class CPRSystem : EntitySystem
     [Dependency] private readonly RottingSystem _rottingSystem = default!;
     [Dependency] private readonly InventorySystem _inventory = default!;
     [Dependency] private readonly AudioSystem _audio = default!;
+    [Dependency] private readonly SharedTransformSystem _transform = default!; // Omu
 
     public override void Initialize()
     {
@@ -90,7 +92,35 @@ public sealed class CPRSystem : EntitySystem
         }
 
         if (!_ingestionSystem.HasMouthAvailable(performer, performer) || !_ingestionSystem.HasMouthAvailable(performer, target)) // Omu, swap parameters to correctly check if target is wearing a blocker
+        {
+            // Omu, fixes the ingestion blocker text not appearing.
+            // Yes, this is shitcode. I am sorry but I don't know how to do this better.
+
+            // first, check if the range requirement is why the interaction failed
+            if (!_transform.GetMapCoordinates(performer).InRange(_transform.GetMapCoordinates(target), IngestionSystem.MaxFeedDistance))
+            {
+                _popupSystem.PopupEntity(Loc.GetString("interaction-system-user-interaction-cannot-reach"), performer, performer);
+                return;
+            }
+
+            // if not range, then check if it's because someone can't do ingesting
+            var attempt = new IngestionAttemptEvent(IngestionSystem.DefaultFlags);
+            if(!_ingestionSystem.HasMouthAvailable(performer, performer)) // first check if the performer has a mask
+            {
+                RaiseLocalEvent(performer, ref attempt);
+                if (attempt.Blocker != null)
+                    _popupSystem.PopupEntity(Loc.GetString("ingestion-remove-mask", ("entity", attempt.Blocker.Value)), performer, performer);
+
+            } else if (!_ingestionSystem.HasMouthAvailable(performer, target)) // if not, check if the target has a mask; this is to prevent mixing textboxes
+            {
+                RaiseLocalEvent(target, ref attempt);
+                if (attempt.Blocker != null)
+                    _popupSystem.PopupEntity(Loc.GetString("ingestion-remove-mask", ("entity", attempt.Blocker.Value)), performer, performer);
+            }
             return;
+            // Omu end
+        }
+
 
         _popupSystem.PopupEntity(Loc.GetString("cpr-start-second-person", ("target", target)), target, performer);
         _popupSystem.PopupEntity(Loc.GetString("cpr-start-second-person-patient", ("user", performer)), target, target);
--- a/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
+++ b/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
@@ -89,7 +89,7 @@ public sealed class CPRSystem : EntitySystem
             return;
         }
 
-        if (!_ingestionSystem.HasMouthAvailable(performer, performer) || !_ingestionSystem.HasMouthAvailable(target, performer))
+        if (!_ingestionSystem.HasMouthAvailable(performer, performer) || !_ingestionSystem.HasMouthAvailable(performer, target)) // Omu, swap parameters to correctly check if target is wearing a blocker
             return;
 
         _popupSystem.PopupEntity(Loc.GetString("cpr-start-second-person", ("target", target)), target, performer);
--- a/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
+++ b/Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
@@ -71,44 +71,7 @@ public sealed class CPRSystem : EntitySystem
 
     private void StartCPR(Entity<CPRTrainingComponent> performer, EntityUid target)
     {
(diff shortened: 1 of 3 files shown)


Willfullyheretical:
Want to be evil.... If the patient has toxic give a minor amount of toxic per do-after to the person performing cpr

codeduck101:
hang on, un-shit-coding the logic behind the CPR fail messages

codeduck101:
> Want to be evil.... If the patient has toxic give a minor amount of toxic per do-after to the person performing cpr

out of scope

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.CodeAnalysis;
```
codeduck101: i genuinely do not know what the diff here is talking about
it looks the fucking same

codeduck101:
un-shitcoded my PR it's good now probably

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
```diff
@@ -93,6 +100,35 @@ public bool HasMouthAvailable(EntityUid user, EntityUid target, SlotFlags flags)
         return false;
     }
 
+    /// <inheritdoc cref="HasMouthAvailable(EntityUid, EntityUid, SlotFlags)"/>
+    /// Omu, allows for taking the reason why the interaction failed
+    /// "message" will be an empty string if the interaction did not fail due to distance or a blocker
+    public bool HasMouthAvailable(EntityUid user, EntityUid target, SlotFlags flags, out string message)
```
QuillTheCreature: Please update the `out string message` to `out string? message`.

C# has nullable variables and we should use them as they provide clearer warnings/errors when used wrongly.



```suggestion
    public bool HasMouthAvailable(EntityUid user, EntityUid target, SlotFlags flags, out string? message)
```

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
```diff
@@ -93,6 +100,35 @@ public bool HasMouthAvailable(EntityUid user, EntityUid target, SlotFlags flags)
...
+    {
+        message = ""; // leave the message empty if nothing is stopping you
```
QuillTheCreature: With the above change we can update this to `message = null` now.


```suggestion
        message = null; // leave the message empty if nothing is stopping you
```

Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
```diff
@@ -89,8 +89,26 @@ private void StartCPR(Entity<CPRTrainingComponent> performer, EntityUid target)
             return;
         }
 
-        if (!_ingestionSystem.HasMouthAvailable(performer, performer) || !_ingestionSystem.HasMouthAvailable(target, performer))
+        // Omu, fix CPR not showing a message for being out of range or for having a blocker
+        // Separated into two checks so that the out variables don't end up unassigned
+        if (!_ingestionSystem.HasMouthAvailable(performer, performer, out string performerMessage))
+        { 
+            if (performerMessage.Length > 0) // Check if message is an empty string; if it isn't, then report it via popup
```
QuillTheCreature: And then any length checks can be changed to proper nullable checks:



```suggestion
            if (!string.IsNullOrEmpty(performerMessage)) // Report it via popup if message is not null
```

Content.Server/_EinsteinEngines/Medical/CPR/CPRSystem.cs
```diff
@@ -89,8 +89,26 @@ private void StartCPR(Entity<CPRTrainingComponent> performer, EntityUid target)
             return;
         }
 
-        if (!_ingestionSystem.HasMouthAvailable(performer, performer) || !_ingestionSystem.HasMouthAvailable(target, performer))
+        // Omu, fix CPR not showing a message for being out of range or for having a blocker
+        // Separated into two checks so that the out variables don't end up unassigned
+        if (!_ingestionSystem.HasMouthAvailable(performer, performer, out string performerMessage))
```
QuillTheCreature: Oh and this needs to be changed too:

```suggestion
        if (!_ingestionSystem.HasMouthAvailable(performer, performer, out string? performerMessage))
```

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
```diff
@@ -70,6 +70,13 @@ public bool HasMouthAvailable(EntityUid user, EntityUid target)
         return HasMouthAvailable(user, target, DefaultFlags);
     }
 
+    /// <inheritdoc cref="HasMouthAvailable(EntityUid, EntityUid)"/>
+    /// Omu, option to take the message for use in server-side logic
+    public bool HasMouthAvailable(EntityUid user, EntityUid target, out string message)
```
QuillTheCreature: Same as the other:

```suggestion
    public bool HasMouthAvailable(EntityUid user, EntityUid target, out string? message)
```

QuillTheCreature (changes requested):

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
codeduck101: shiiiit i tried this last time but it was complaining about how you can't have nullable out
let me see if i did it wrong the first time

codeduck101:
changes addressed in latest push

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
codeduck101: seems to work and i just fucked it the first time around

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
codeduck101: if anyone can undo this i'd be very happy

codeduck101:
boss i refactored the whole fucking thing to avoid checking twice for reason and message
plus i added that messages show up if you're interrupted mid-CPR
please see new video

codeduck101:
i'd like to note that i am very happy with what i have at present and i feel that it is ready for review

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
ishkab: Your original commit removed the byte order mark at the beginning of this file. I have re-added it in commit 9b95ebf1f70c81162d2e74135f11abd3297ff13e. You can modify this using the status bar at the bottom right of most good editors: it's normally included in the file's encoding settings. See VS Code example below.

Image

See also this helpful wiki article: https://en.wikipedia.org/wiki/Byte_order_mark

ishkab (approved):

Content.Shared/Nutrition/EntitySystems/IngestionSystem.API.cs
QuillTheCreature: Why is it even in BOM in the first place. 😮‍💨
Good catch though.

QuillTheCreature (approved):

codeduck101:
i do not know why this is yet to be merged
is that a button you have to separately press

PR #1318, opened 2026-07-28
Allow dead players to respawn as a new character

MarkerWicker:
(paging @CliveOcelot for this)
## About the PR
This PR allows players to respawn as a new character once they've been ghosted. Logic is ported from WD.

## Why / Balance
This branch has been sitting for a while, but I've opened this MR at the request of cyg. Go ask her for the balance justification.

## Technical details
Minorly edited a few upstream files. Added functions to allow returning to round.

It checks your last char's name against the new name, if it is 85% or more similar then it rejects the attempt, and alerts the player, if it is below 85% but above 50% then it alerts admins instead, while allowing the player to join.

Note, the character name similarity check is disabled if the person is marked as an active admin (readmined, deadmined does not count) 

The GhostRespawnTime cvar is how many minutes a player must be dead for before being allowed to respawn, this defaults to 15.

By default the GhostRespawnMaxPlayers cvar is set to 40, if there is more than this number of people on the server (note, total players not just alive/inround players) it will not allow respawning.

## Media
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- add: Added respawn button


Content.Client/Ghost/GhostSystem.cs
```diff
@@ -41,6 +41,7 @@
 using Content.Client.Movement.Systems;
 using Content.Shared.Actions;
 using Content.Shared.Ghost;
+using Content.Shared._White.Ghost; // Omustation - Port WD Respawn Button
```
NotActuallyMarty: fuck marking usings
```suggestion
using Content.Shared._White.Ghost;
```

Content.Server/GameTicking/GameTicker.Spawning.cs
```diff
@@ -558,6 +575,81 @@ public void SpawnObserver(ICommonSession player)
                 $"{player.Name} late joined the round as an Observer with {ToPrettyString(ghost):entity}.");
         }
 
+
+        // begin Omustation - Port WD Respawn Button
+        #region WD Ghost Respawn
+
+        /// <summary>
```
NotActuallyMarty: did not read code but i would prefer whatever is going on here be in its own system rather than inserting a massive region into upstream files

NotActuallyMarty (changes requested):
did not read code im on mobile but  whatever is going on just needs to be moved into its own system for organization sake

also using a chat message is bad and missable granted ive not tested ingame but would prefer either a ghostrole selection style warn popup or port starlight's respawn system instead which had a lot of ui shit from what i remember

CliveOcelot:
note for people testing this, you need to set game.lobbyenabled to True if testing on debug

PR #1328, opened 2026-07-30
Make binary a radio again

CliveOcelot:
## About the PR
Makes binary back into a radio instead of a hivemind.

## Why / Balance
So that damage can garble the voice, alongside allowing radio jammers to work on it. It also means borgs/AI can state laws over binary again. 

## Technical details
Binary encryption keys have been returned to the syndicate uplink in place of the implanter, same goes for in the hacker bundle.

RD's hud no longer provides access to binary but still acts as a translator for robottalk, instead they have a binary encryption key in their locker.

## Requirements
- [X] I have read and agree to the Contributor License Agreement.

## Breaking changes
Anything attempting to use the binary hivemind will now fail, this is intentional so to force people to use the radio instead.

**Changelog**
:cl:
- tweak: Changed Binary back into being a radio channel, it can be used on :b


SirCannonFodder:
The radio jammer working on binary, that I get. 
But the speech degradation effecting binary is going to lead to more situations of cyborgs effectively being round removed because they're unable to communicate the information needed for them to be recovered.

CliveOcelot:
> The radio jammer working on binary, that I get. But the speech degradation effecting binary is going to lead to more situations of cyborgs effectively being round removed because they're unable to communicate the information needed for them to be recovered.

So? if i crit a person in maints they cant talk at all, and its far harder to revive a dead person than a cyborg, you can make an entirely new cyborg in under 60s in science, hell some people automated it, by SOP sci is required to have a posi ready at all times, so even if they do get fully RRed, they can just ghostrole into a new borg.

Resources/Prototypes/Entities/Mobs/Player/mothershipcore.yml
```diff
@@ -144,12 +144,13 @@
...
     - Xenoborg
-    #- Binary Goob
+    - Binary # Omu, make binary a radio again.
   - type: ActiveRadio
     channels:
     - Mothership
     - Xenoborg
-    #- Binary Goob
+    - Binary # Omu, make binary a radio again.
+  - type: XenoborgMothership
```
CliveOcelot:
```suggestion
```

Thesia000 (approved):

feistyfedora (approved):

PR #1329, opened 2026-07-30
[DRAFT] Port Funky's Gun Lag Compensation

codeduck101:
## About the PR

port of funkystation's gun lag compensation and its fixes

## Technical details

port of funky's
https://github.com/funky-station/funky-station/pull/1833 (merged in with twelve billion SPDX conflicts and several with goobstation and shitmed and i think woundmed?)

which mentions that RMC's doesn't deal with energy weapons (or bows?) at
https://github.com/funky-station/funky-station/pull/1901 (merged)
https://github.com/funky-station/funky-station/pull/1916 (merged)

these seem related
https://github.com/funky-station/funky-station/pull/1927 (merged)
https://github.com/funky-station/funky-station/pull/1994 (merged)

we have conflicting names so i had to change PhysicsSystem to Robust.Client.Physics.PhysicsSystem and C.Sh.FixedPoint to C.Goobstation.Maths.Fixedpoint

so here's how it is so far (2026/7/31)
*  bows are cooked in general: if you try to re-use an arrow by pulling it out of someone then reinserting it into your bow, i'm preeeeeeetty sure there's a client-side crash, and you just cannot shoot or hit anything with the reused arrow (it bounces around w/o gravity)
* syringe guns are fucked: the syringe is set at a strange angle and has a firing delay like lasers. they also only inject 4 ticks of reagent before stopping and cannot be fired twice, just like bows. regarding the angle, i believe it's because the syringes aren't using the ThrowingAngleComponent and are being hyucked at 45 degrees off (the needle is not facing the travel direction)
*  i'm pretty sure all burst-fire weapons are cooked client-side since i think i'm missing some goob-specific checks for client-side burst fire, but at least the drozd performs as expected on server
  * the drozd shoots one client-side bullet, then the rest of the bullets follow up when the server realizes it should shoot. also it client-side full autos if you hold down shoot, and the "gun unbolted" popup shows up as if you were mashing every frame. again, completely fine on server but is really goofy on client
  *  2026/8/1: also, the drozd can somehow split its SERVER-SIDE bullet bursts by disabling harm mode mid-burst, then re-enabling it at any time in the future
  * the abielle does roughly the same thing
  * it seems that when unloaded and holding down shoot, burst-fire weapons attempt to fire `some number of` times per burst: the drozd dry-fires 6 times and the abielle dry-fires 11 times
* lecter, energy shotgun and disablers are like kinda half cooked where they show both the client predict and the server shot
  * the atreides (didn't know that was a gun) has very intriguing client-side ammo counter issues when you tapfire it. it's a full-auto gun though so i don't exactly know why it's doing this (it attempts to shoot a client-side three-shot burst)
  * the lecter attempts to fire a client-side two-shot burst. i think this might be an issue with all non-spread guns
* the kammerer and enforcer actually work really well but from what i can tell, the client's prediction has no actual tie to the server's bullets (so the client and server will often disagree on mildly elevated ping). additionally, the client-side bullets have different properties (see lower down) but the server-side performance is the same
  *  2026/8/1: the L6C ROW (a borg weapon), the anaconda, and the water pistol (and i assume spray nozzle + backpack water tank) work the same way i think it's because it has no magazine and spawns the projectile and empty casing itself rather than having the cartridge spawn a client/server projectile
  *  the enforcer spawns its shell and server-side projectiles according to the server, and the client/server bullets do slightly mismatch, but it does not have the issue where it shoots two bullets per bullet
* hitscan lasers are also half-cooked where the red flash on the target appears instantly (without a laser), but then the actual server-side laser shows up and causes the flash again
  *  actually this happens for all of the non-shotgun weapons. the client bullet makes a hit noise and then the server bullet makes a hit noise
*  for whatever reason, all non-shotgun guns i tested have a client-side WHOLE-ASS SEPARATE PROJECTILE that has different properties compared to the real server-side projectile, meaning that the speed and sprite of the projectile is different on client compared to server
  *  update 2026/7/31: not a totally separate projectile. there is linear dampening on for ALL client-side projectiles that respects gravity (as in that the projectile will slow down if under gravity). unsure if this is also true for server-side projectiles, but projectiles are marked to obey gravity so i dunno
  * update 20206/8/1: client-side projectiles slowing down isn't based on gravity but rather being over a grid. definitely not true for server-side projectiles, as they're meant to have no dampening or friction
* even when setting the cvar to disable client-side prediction, you can still see a server-side and a client-side muzzle flash
* the ammo counter on revolvers is minorly bugged where it briefly flashes one bullet ahead if you dry fire it
* the PKA and PK shotgun kinda work? (minus the ammo counter/reloading visuals desync) the proto-kinetic repeater shoots four times client-side but only reloads three times, and the pistols have the same issue as the lecter where it shoots two bullets 
* the antique laser pistol can sometimes fire a server-side projectile but client-side consider the gun still "out of ammo" and fail to play the firing noise

## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an ingame showcase.

- [ ] I have read and agree to the Contributor License Agreement.

## Breaking changes

all prototypes in `projectiles.yml` MUST NOT have a `fly-by: *flybyfixture` attribute or it will cause the whole file to fail to load

**Changelog**


codeduck101:
so here's how it is so far:
*  the drozd is cooked client-side since i think i'm missing some goob-specific checks for burst fire but performs as expected on server
* lecter, energy shotgun and disablers are like kinda half cooked where they show both the client predict and the server shot
* the kammerer and enforcer actually work really well but from what i can tell, the client's prediction has no actual tie to the server's bullets (so the client and server will often disagree on mildly elevated ping)
* hitscan lasers are also half-cooked where the red flash on the target appears instantly (without a laser), but then the actual server-side laser shows up and causes the flash again
  *  actually this happens for all of the non-shotgun weapons. the client bullet makes a hit noise and then the server bullet makes a hit noise
*  for whatever reason, all non-shotgun guns i tested have a client-side WHOLE-ASS SEPARATE PROJECTILE that has different properties compared to the real server-side projectile, meaning that the speed and sprite of the projectile is different on client compared to server

Anui-ElXx:
chat, is this what the kids refer to as "peak" this days?

Anui-ElXx:
peak

codeduck101:
closing due to skill issue and my entrance into college
related PR may be made by a more skilled developer than i am

PR #1335, opened 2026-08-01
Escaped special characters in player last words in Round End Summar Window

QuillTheCreature:
## About the PR
Escaped special characters in player last words in the round end summary window.

This should prevent the crash which just happened in the last round. (Round 6194)

## Why / Balance
No round end summary window is bad.

## Technical details
Someone had some invalid markdown formatting as their last message (probably a `[` or something like that) which resulted in a crash, so instead let's escape that message.

## Media
image

## Requirements

**Changelog**

:cl: Quill
- fix: Fixed round end summary window crashing when a player writes a badly formatted message as last message.


feistyfedora (approved):
Looks good to me

CliveOcelot (approved):

PR #1336, opened 2026-08-01
Removed RemoveMarkupOrThrow in ChatSystem to allow invalid markdown from being set as chat messages

QuillTheCreature:
## About the PR
Sending a message like `[` would throw an exception and not allow the message to be sent. This is a PR to allow messages containing invalid markdown. 

## Why / Balance
Let creatures send their messages instead of not doing anything when hitting enter.

## Technical details
Before sending a chat message, there's a `FormattedMessage.RemoveMarkupOrThrow(originalMessage);` which should strip the message of all formatting.

However, since this a `OrThrow()` message, this means that the method crashes when a creature sends an "invalid formatted" message, like `Hello [`.

We decided that the player should be able to actually send what they type so we removed that line.

## Media
image

## Requirements

**Changelog**

:cl: Quill
- fix: Fixed chat messages from not sending because of invalid markdown tags.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -758,10 +758,18 @@ public sealed partial class ChatSystem : SharedChatSystem
             return;
 
         // The Original Message [-] Einstein Engines - Language
-        // Omu - RemoveMarkupOrThrow() ... throws an error when it's a invalid message, resulting in the message not appearing in chat.
-        // This could happen when for example someone sends a `[` or other things.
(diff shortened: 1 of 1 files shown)


CliveOcelot (approved):

NotActuallyMarty (changes requested):
I would still preffer this used `RemoveMarkupPermissive` as that way we remove any chat that has markup in it i.e.

Image

turns to:

Image


so tldr;

Image

feistyfedora (approved):
It's a little funny that this also can throw an exception
Looks okay to me

NotActuallyMarty (approved):

PR #1342, opened 2026-08-02
Give xenos structural, make barricades unable to be climbed

CliveOcelot:
## About the PR
Some changes that should hopefully make xenos a bit better to play, I would have prefered to give them the acid action back but its very broken now, and I was not able to fix it easily, so this will do.

Barricades loose the ability to be climbed / crawled over, making them actually function for defensive lines, given that you can instantly toggle the gate parts, this shouldn't be a major balance issue, but rather makes them more useful and interesting.

## Why / Balance
Xenos needed a way to break through things, and barricades needed to actually stop people. done in the same pr as to not break ballance while the other one is not merged. (Tested xenos against the barricades and they can break through them with relative ease).

## Technical details
Adds MidImpassable to the barricades, and removes TableLayer, preventing crawling over, then removes Climbable so they cannot be climbed onto.

All adult xenos now deal a base of 20 structural, with Praetorians doing 30, and the Queen doing 40.

## Requirements
- [X] I have read and agree to the Contributor License Agreement.

**Changelog**
:cl:
- tweak: Barricades cannot be climbed or crawled over any more.
- tweak: Xenos now deal structural (20 for adult xenos, 30 for Praetorians, 40 for Queens).

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Structures/Walls/barricades.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Structures/Walls/barricades.yml
@@ -16,13 +16,14 @@
         mask:
         - FullTileMask
         layer:
-        - TableLayer
+        #- TableLayer # Omu
+        - MidImpassable # Omu
         - LowImpassable # no mice or crawling chuds allowed, its flush with the floor
   - type: RotationDrawDepth
     southDrawDepth: OverMobs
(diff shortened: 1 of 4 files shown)


feistyfedora (approved):
Looks good to me

Thesia000 (approved):

PR #1350, opened 2026-08-04
The great bioluminescence update

QuillTheCreature:
## About the PR
Added a "glowing" checkbox to every color in the marking system.

All characters **must** glow.

This is a partial modified port of https://github.com/ss14Starlight/space-station-14/pull/641

I decided against porting the eye/hair/facial hair changes right now as those require a database change and I wanted to port the simpler thing first.

## Why / Balance
Currently only specific markings were allowed to glow, which kind of sucks and doesn't make a lot sense why *this one* tattoo could glow, but the other one could not.

## Technical details
I added a `glowyBits` bit field to the marking system. This is a bit field which indiciates which color index of a marking is either glowing or not.

I decided to use a bit field as it is similar to the `bool` variable of the starlight system and a lot simpler to handle than a `List<bool>`.

Then I upgraded all references in the humanoid appearance system, wizard system, and changeling system.

And then I added checkboxes to the single marking picker and marking picker to support that change.

## Media

image

image

## Requirements

**Changelog**

:cl: Quill
- add: Added per color glowing toggles to markings.


CliveOcelot (approved):
tested, works:

Image

ishkab (approved):

PR #1354, opened 2026-08-04
Add nodecrawling from DV

harryhulse12-alt:
## About the PR
Ports nodecrawling from DV and renames old tilemovement to oldtilemovement - is still used in a lot of places.

## Why / Balance
Cleaner vent crawl

## Technical details
Cherry-pick and manual port from DV

## Requirements

## Breaking changes
TileMovementComponent was renamed to OldTileMovementComponent

**Changelog**
:cl:
- tweak: Changed ventcrawling to a new system from DV, it should be more stable to use now.


CliveOcelot (changes requested):
Remove any reference to `- type: VentCrawler` anything that could ventcrawl should be using nodecrawl instead

CliveOcelot:
this still crashes, ping me when fixed

Content.Goobstation.Shared/Vehicles/SharedVehicleSystem.cs
```diff
@@ -235,8 +235,8 @@ private void Mount(EntityUid driver, EntityUid vehicle)
     {
         _mover.SetRelay(driver, vehicle);
 
-        if (HasComp<TileMovementComponent>(driver))
-            EnsureComp<TileMovementComponent>(vehicle);
+        if (HasComp<OldTileMovementComponent>(driver))
```
MarkerWicker: Why rename so many instances of TileMovementComponent in upstream files to OldTileMovementComponent? Wouldn't it be better to leave them as-is and instead refer to the new system "NewTileMovement" in order to prevent future upstream merge conflicts?

harryhulse12-alt: Yeah, very good point, changed to that

PR #1379, opened 2026-08-08
Fix Cryopods from crashing the server when built due to shitcode

NotActuallyMarty:
Something something i fixed this on goob months ago with shitmed fixes and now some fuckass system that runs on an insert event checks an ID before it is assigned and occasionally crashes.

im self merging i don't care this crashes servers.


4Hangar:
SPARK STATION IS SAVED

PR #1395, opened 2026-08-11
Make Aloxadone Heal Cold Damage

codeduck101:
## About the PR

buffed alox to heal 10 cold too (it's implied that alox should be able to heal frostbite and there isn't a whole lot of reason why it shouldn't heal 10 all-heat instead of 10 heat/shock/caustic. like if it heals 10 caustic it better heal cold)

i dunno why it didn't heal this in the first place really

## Why / Balance

seems like a major oversight with aloxadone to not heal cold damage. plus, it'll help with un-obliterating station pets who walk out into glacier and die

## Technical details

one-line YAML change in wizden `Prototypes\Reagents\medicine.yml`

## Media

Screenshot 2026-08-11 144513

i verify that i tested with a real cryo pod and it does in fact heal cold

## Requirements

## Breaking changes

**Changelog**

:cl: Duuckie
- add: Aloxadone now heals cold damage as implied by its description.

### DIFF
--- a/Resources/Prototypes/Reagents/medicine.yml
+++ b/Resources/Prototypes/Reagents/medicine.yml
@@ -1778,6 +1778,7 @@
             Heat: -10.0
             Shock: -10.0
             Caustic: -10.0
+            Cold: -10.0 # Omu, fix oversight with Aloxadone not healing cold damage
 
 - type: reagent
   id : Mannitol # currently this is just a way to create psicodine


Floofbutt:
Imho cryotube still heals all species too slow to be remotely useful compared to chems. 

Temperature threshold should be raised across the board on all cryochems, or the rate that the tube alters species' temperature's should be accelerated, if you want to see more tube usage. 

Also station pets still have the same cooling issues vulps had.

codeduck101:
the cryo tube is slow is quite true, but it is meant to be a "set and remember in 2 minutes" kind of thing instead of normal chems which are intended for a quick turnaround

cryo is cheap and you can afford to waste cryox and dox


also that's extremely out of scope for what this PR is doing but i'd consider it if i knew what the discussion around cryo was like

Toaster-Box (approved):

feistyfedora (approved):
Agree it seems like an oversight. This aligns with current chem description.

pinktanuki (approved):
Looks good

PR #1415, opened 2026-08-15
Make secure weapon crates actually secure

CliveOcelot:
## About the PR
Secure weapon crates (and the crates that are children of it ex: enforcer crate) now until first unlock will protect their contents even when destroyed, aka: if you use a exosuit drill on a shotgun crate without first unlocking it, all contents will also be destroyed. after first unlock this is disabled by means of removing a tag from the crate.

## Why / Balance
So if cargo wants to arm they at least need to get someone with armory access (or someone with an amag) to open the crates, hopefully should decrease cargonia related incidents. 

## Technical details
I'm not exactly pleased with how I did this but it works and it was the best method I could think of.

CrateWeaponSecure now has a new behaviour placed before the DoActsBehavior of destruction "DeleteAllLockedContainerContentsBehaviour", this behavior when triggered by the crate being broken, checks for the tag "NotOpenedSecureCrate" on the crate, if its found then the contents are destroyed along with the crate, with this tag being removed when the crate is unlocked by any means (and is not re-added upon crate being locked again). if the tag is not found then the crate will break, but the contents will drop as usual.

## Media
https://github.com/user-attachments/assets/d9e5daec-4a39-4cf2-8bc6-64eaf6ca60ab

## Requirements

**Changelog**
:cl:
- tweak: Secure weapon crates, such as those from cargo, now will destroy their contents if broken open before being unlocked.


raincally:
i think this will overcomplicate crew arming during stuff like nukies or zombies where someone with armory access Probably wont be there

codeduck101:
woe, armory access be upon the QM

Willfullyheretical:
I like it, though it makes me think it would be cool to add a buzzer to crate on hit or some kind of melt effect to destroyed crates that have not been unlocked before

LoneSarge:
I get the intention here but I think this really makes crew arm a LOT more unnecessarily complicated and prone to just wasting cargo's resources

Thesia000 (approved):

feistyfedora (approved):

PR #1419, opened 2026-08-16
Chemistry: Distillation

ishkab:
## About the PR
This PR introduces a "quick and dirty" implementation of distillation, a new mechanic in the chemistry lab. Certain reactions now produce an impure variant of the original chemical. These impure mixes are diluted with impurities leftover from the reaction and are generally worse off for it. You can filter out the impurities using a distilling setup, a new station for reacting chemicals.

## Why / Balance
¯\\\_(ツ)\_/¯ It seemed like a good change relative to how difficult it was to implement.

## Technical details
It's all YAML. Can you believe it?

## Media
I'm not putting media yet because my spritework sucks.

## Requirements

## Breaking changes
Very few anticipated.

**Changelog**

:cl:
- add: Distilling setup.
- tweak: Some chems now should really be distilled before you use them.


Resources/Prototypes/_Omu/Reagents/impure.yml
```diff
@@ -0,0 +1,106 @@
...
+        damage: { types: { Poison: 4 }}
+      - !type:HealthChange # OD >= 35; higher because of impurities
+        damage: { groups: { Burn: 4 }} # same as pure
+        conditions:
+        - !type:ReagentThreshold
+          min: 35
+        - *yow
+
+- type: reagent
+  id: ImpurePenteticAcid
```
Toaster-Box, line 82: suggestion, make impure pent reduce your blood levels

Resources/Prototypes/_Omu/Recipes/Reactions/distillations.yml
```diff
@@ -0,0 +1,27 @@
...
+  id: DistillImpureSalicylicAcid
+  requiredMixerCategories: *d
+  reactants: {ImpureSalicylicAcid: 1}
+  products: {SalicylicAcid: 0.65, Impurities: 0.35}
+
+- type: reaction
+  id: DistillImpureOxandrolone
+  requiredMixerCategories: *d
+  reactants: {ImpureOxandrolone: 1}
+  products: {Oxandrolone: 0.42, Impurities: 0.58}
```
PhantomDergwulf, line 21: This is a little excessive for oxan. if i understand this correctly, you're only making 42% of the original oxan? so wouldn't that require you to make over DOUBLE the amount of oxan you would normally? I'd say bump it up to at least 50% or above personally.

PhantomDergwulf (commented):
I need to put something here to submit the comment uhh. fuck :3

Resources/Prototypes/_Omu/Reagents/impure.yml
ishkab: oo

ishkab: remind me about this later

Resources/Prototypes/_Omu/Recipes/Reactions/distillations.yml
ishkab: My original idea with this ratio was that:
- Oxan is so potent that halving its production probably wouldn't change the overall amount of healing it's doing in a round, assuming it continues to see use.
  - This point covers instances where a full 15u syringe is administered where e.g. a 5u dose would work just as well and save 10u.
  - This point also covers instances where oxan is "misused" to treat a small amount of a single damage type or similar.
- Nerfing oxan in this way opens up the possibility of chemists reprioritizing e.g. the so-called T2 burn chems (pyrazine, leporazine, insuzine, and sigynate) since oxan is "more work" now.
- Having it be 100 -> 42 - as opposed to some nice multiple of 10) makes the reaction icky.

ishkab:
notes for later implementation:
- faster distillation machine for nukie use case, try 30s operation time
- guidebook entry

PR #1467, opened 2026-08-22
Fix Packaging and set self contained

NotActuallyMarty:
good one upstreamer. Also im setting self contained so i dont have to install dotnet10


CerberusWolfie (changes requested):
You forgot a ”.

PR #1554, opened 2026-08-25
Port Secure Command Terminal from Starlight

CliveOcelot:
## About the PR
Ports the secure command terminal from starlight, which integrates into the primary bridge comms console, alongside keycard readers that are to be mapped in head offices (for now dep comms consoles also function as key card readers, but this is to be removed later down the line).

This terminal has a list of requests it can start votes for, with varying numbers of votes from varying personnel required, current requests vary from EAing every door, or EAing maints, to a request for code Gamma, or even the nuke codes.

Requests support requiring admin approval, and in the absence of one, can be set to bypass approval if wanted for that specific request.

This PR also ports some ships/drop pods used by the system, alongside Starlights AlertArmory system, which is also used by it.

Notably, it differs from starlight as it draws funds from a set account, currently all requests are set to use cargo's account, but this can be set on a per rule basis.

This PR also makes it so that enacting Code Amber requires using the secure terminal system, with authorisation required from either Captain, Hos, or NTR for both votes.

## Why / Balance
Its a cool system and can be used for future things.

Balance wise I do wish to swap the drop pods etc to use gear that isn't too strong for the threat they would be called in for, but I require other's input on this.

## Technical details
Port of quite a few starlight prs, and some partial ports of dependencies that it requires.

## Media
image
image

### Shuttles / Drop pods

Emergency Medical Pod:
image

Emergency Gamma Armory:
image

ERT Chaplain:
image

CBRN:
image

ERT Engi:
image

ERT - Full:
image

ERT Jani:
image

ERT Med:
image

ERT Sec:
image

Warops Droppod:
image

## Requirements

**Changelog**
:cl:
- add: Ported the Secure Command Terminal from starlight, read its guidebook page for more info.
- tweak: Code Amber now requires a vote through the Secure command Terminal.


CliveOcelot:
Marking this as do-not merge till I get all tests passed and work out something decent balance wise for the gearing.

CliveOcelot:
codewise, now works, just need to sort out balance of the ERT roles it uses

CliveOcelot:
ERT roles updated to use Blue alert for the majority (Exceptions are ERT Security, ERT full, and CBRN (this does not have blue alert versions nor does it make sense to have.)

Blue alert protos updated to be more in line with ERT equipment

PR #1610, opened 2026-08-31
Visually impaired characters listening to whispers fix

redmagicspider:
## About the PR

Altered a part of the chat system to fix with how characters with impaired visions would view whispers and messages in sign language.

## Why / Balance

Before this, characters that had vision impairment would be unable to hear any whispers and any characters that doesn't have the 'Blind' trait (mantle beasts or short-sighteds) couldn't see sign language even though it would make sense for them.

## Technical details

DeafnessBlindnessSystem.cs in Content.Goobstation.Server/Chat has the 'PermanentBlindnessComponent' hooked on OnBlindnessOverrideInRange method altered, checks for 'RequiresSight' then checks if the entity in question is completely blind, if true then it cancels the message on client.

ChatSystem.cs has slight alteration due to how the deafness/blindness check works in 'SendEntityWhisper', it borrows from the method found in 'SendVoiceInRange'... I would've kept the original code but I don't have any experience with the SS14 engine to be able to just extract the 'Blindness' variable from the component and keep the condition, so I had to do an event call.

## Media

Warning for the moving effects from the blurry vision shader intensified by the video compression...

https://github.com/user-attachments/assets/420e22f7-8b68-4d52-8fd9-406abca1d23a

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed blind/visually impaired characters to now hear whispers.
- fix: Fixed visually impaired (not totally blind) characters to see signs.

### DIFF
--- a/Content.Goobstation.Server/Chat/DeafnessBlindnessSystem.cs
+++ b/Content.Goobstation.Server/Chat/DeafnessBlindnessSystem.cs
@@ -54,6 +54,8 @@ public sealed class DeafnessBlindnessSystem : EntitySystem
     {
         if (!args.RequiresSight)
             return;
-        args.Cancel();
+        if (ent.Comp.Blindness <= 0)
+            args.Cancel();
+        // Signs are completely disabled if the entity is totally blind.
     }
 }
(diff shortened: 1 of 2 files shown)


redmagicspider:
i am not good with computers
it's fucking midnight

PR #1624, opened 2026-09-01
Earlymerge nuke bad lathe

NotActuallyMarty:
See lathe PR upstream.

https://github.com/Goob-Station/Goob-Station/pull/7107

Sorry upstreamer. Skip the lathe PR forme thanks. We squashed an upstream anyway so like, its fucking joever.

:cl:
- fix: Ore lathes now print raw materials properly into storage in bulk instead of having to print one by one.


harryhulse12-alt (approved):
Yes Marty boss

PhosphateWithManyOs (approved):

PR #1630, opened 2026-09-02
Check all station report papers

QuillTheCreature:
## About the PR

This PR updates the station report system to check all station report papers on the station instead of only the first one.

I'm really not sure but from looking at how the station reports gets assembled it looks like this might be the issue that sometimes the station report doesn't show up in the end of round report?

## Why / Balance

Hopefully a bugfix

## Technical details

Instead of checking only the first station report I've updated the system to check for all station reports and pick the first one where the text isn't the default one.

## Requirements

**Changelog**
:cl: Quill
- fix: Fixed a bug with station reports not showing up when there are multiple reports on the station.


harryhulse12-alt (approved):
Tested, report sends fine
even if this does break something

It literally can't be worse than the station report just not loading

Image

PhosphateWithManyOs (approved):

PR #1633, opened 2026-09-02
Balance changes for a decent few things

CliveOcelot:
## About the PR
Balance changes for a decent few things, see below for details.

## Why / Balance
Removes AP from .30 (Its called LightRifle why would it have AP, its also fired by the L6 and this is overdue), still a good ammo type without AP.

Re-namespaces some of the APDS stuff to be in the correct place, makes its research also require nonlethal ammo research.
APDS is now printed a shell at a time, is more expensive per shell, and also requires plasteel.
APDS now deals less damage (10 peirce 20 struct) but retains full AP.

Colossus gets one extra tile of melee range, people were able to grab it, preventing it from moving properly, and would be outside of its reach. this allows it to still hit them (Also discourages people from meleeing the giant melee creature)

R6 Shotgun fire rate decreased to 2.25, still higher than enforcer, it also received a reduced spread as well.

## Requirements

**Changelog**
:cl:
- tweak: .30 Rifle no longer has AP.
- tweak: APDS shells are now more expensive, and deal less damage, but still retain full AP.
- tweak: Colossus gets an extra tile of melee range to prevent people from grabbing it and making it unable to move properly.
- tweak: R6 Shotgun fire rate decreased, and spread decreased a small amount.


harryhulse12-alt (approved):
cool beans

Toaster-Box (approved):

PR #1711, opened 2026-09-13
funny penman ammo pr

raincally:
## About the PR

Add ammo to uplink and hacked lathe
## Why / Balance

Added penman mags to emagged autolathe and uplink
## Technical details

yml
## Requirements

## Breaking changes

**Changelog**

:cl: raincall
- fix: Added LRR-30 penman magazines to the uplink and emagged autolathes

### DIFF
--- a/Resources/Locale/en-US/_Omu/store/uplink-catalog.ftl
+++ b/Resources/Locale/en-US/_Omu/store/uplink-catalog.ftl
@@ -41,6 +41,9 @@ uplink-hobbyist-spellbook-desc = A book given out by the Wizard Federation to ge
 uplink-sniper-penman-name = LRR-30 "Penman" DMR
 uplink-sniper-penman-desc = A prized long range product of Waffle Co. able to down targets fast and accurately at medium to long ranges. Chambered in .30 long rifle magnum, fires in semi-auto, has a medium strength optic.
 
+uplink-magazine-longrifle-name = .30 Long Rifle Magnum Magazine
+uplink-magazine-longrifle-desc = A magazine of .30 Long Rifle Magnum, compatible with the LRR-30 Penman.
+
 uplink-ammo-longrifle-name = Box of .30 Long Rifle Magnum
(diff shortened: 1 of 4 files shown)


starsoda743:
please edit the title of the PR to be more informative. it doesn't tell the reader what it does for the ammo

codeduck101:
legally also FTL changes but it's basically all YAML sure
the "why" section is restating the "about" but obviously the "why" is "yeah you can't get any more magazines for the penman except by buying another penman which is dumb"
re-triaging to reflect the current PR state: reclassified as a bug fix because there's no way you shouldn't be able to buy/manufacture more magazines for the penman and it's more than likely going to be a huge problem for the gun's viability without additional magazines

Thesia000 (changes requested):
AP ammo should always require plasteel to print please adress this in the recipy as rn this prints quite a lot of bullets for 1.5 steel that have AP if I am reading this correctly

Marikiii:
> AP ammo should always require plasteel to print please adress this in the recipy as rn this prints quite a lot of bullets for 1.5 steel that have AP if I am reading this correctly

It's roughly the same AP as .20 rifle, just higher damage, it should be fine

Marikiii:
It should cost more steel for sure though

LukaSlade:
So you guys know:
.30 rifle magnum has 35% AP, 
.20 rifle has 15% AP

Also, the round itself isn't specifically an "armor piercing round"; its FMJ like the rest but is just a magnum cartridge meaning it has more powder and thus goes faster, and penetrates more. I don't think it should need plasteel.

LukaSlade:
Also it was technically intended for you to not be able to get more than your 3 extra mags kind of like the Hristov's single mag, so its not TECHNICALLY a bug fix. 

I do agree its better this way though, having magazines separately purchasable and printable makes more sense anyway.

LukaSlade:
Justifying the remark about viability btw so I don't look stupid:

My previous thought process would be that people would keep the three mags in their bag and buy a couple .30 lrm boxes to refill said mags when they get used up. (again, similar to the Hristov but with more mags)

Again not to confuse things i'm 100% on board with this PR

codeduck101:
well ok then

PR #1739, opened 2026-09-16
Rev polish

harryhulse12-alt:
## About the PR

Adds polish to revs pre playtest
Updates guidebook appropriately
Adds morale icons to indicate to command and head revs what progress they are making

Fixes the book targetted not applying the comp
## Why / Balance

I decided to do it instead of relax or play.

## Technical details

Adds faction icons for four mood states - having passed a morale check and being immune to conversion
Losing against morale and being close to conversion
Average
Having high morale and being far from conversion

Makes a new event for the book being targetted - required for it to ensure the morale comp so it works as intended

## Media

image
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added rev morale indicator icons - makes it easier for everyone to know whats going on.


Jerulean (approved):
Seems like a good addition to the revs gamemode

codeduck101:
i feel that the "about" section being limited to "adds polish to revs" is a bit too general since i had to skim the code to find out what tha hell's going on here; as far as i can tell, this PR solely adds morale indicator icons? good shit but i'd like to know if there's anything else going on that i don't see
(it's not like this matters for in-game purposes but i'd like to clarify if this is all that's intended to be added)

also, i think an image is missing from the media showcase: you write that there are four mood states but i only see three images of i assume three states. where is the image for immune to conversion?

Content.Server/_Omu/Revs/MoraleEnsured.cs
```diff
@@ -0,0 +1,68 @@
...
+        var userFactionIcons = EnsureComp<CustomFactionIconsComponent>(uid);
+        userFactionIcons.FactionIcons.Remove(faction);
+        Dirty(uid, userFactionIcons);
+    }
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        if (!_gameTiming.IsFirstTimePredicted)
```
feistyfedora, line 53: This is only on server so I don't see how this could be predicted.

Content.Server/_Omu/Revs/MoraleEnsured.cs
```diff
@@ -0,0 +1,68 @@
...
+    private void OnStartup(EntityUid uid, MoralePassedComponent component, ComponentStartup args)
+    {
+        var userFactionIcons = EnsureComp<CustomFactionIconsComponent>(uid);
+        userFactionIcons.FactionIcons.Add(faction);
+        Dirty(uid, userFactionIcons);
+    }
+
+    private void OnShutdown(EntityUid uid, MoralePassedComponent component, ComponentShutdown args)
+    {
+        var userFactionIcons = EnsureComp<CustomFactionIconsComponent>(uid);
```
feistyfedora: How would we get to the point of shutting down this comp and not having CustomFactionIcons?
It's added on startup, so I don't think it's needed here. 
Why wouldn't we remove the comp here?

Content.Server/_Omu/Revs/MoraleSystem.cs
```diff
@@ -124,16 +149,35 @@ private void OnChange(Entity<MoraleComponent> ent, ref MoraleChangedArgs args)
         ent.Comp.MoraleValue += args.Amount;
 
         var morale = ent.Comp.MoraleValue;
+        string faction;
 
-        if (morale <= 0f)
+        switch (morale)
```
feistyfedora, line 160: Why not change the icons if it's increasing/decreasing instead of at thresholds?
Changing icons based on recovery/morale decreasing would communicate to the player immediately.

feistyfedora (dismissed):
Mainly questions on some logic
Will come back to this later to test in dev

harryhulse12-alt:
Mfw 2 am coding makes me miss things, let me address it all

Content.Server/_Omu/Revs/MoraleSystem.cs
harryhulse12-alt: Cause as it stands its still big bursts of loss and slow gain, it would constantly be green with little indication as to how close they are

Content.Server/_Omu/Revs/MoraleEnsured.cs
harryhulse12-alt: Theres a better way of doing it true, but other systems use customfactionicons and we might not want to just wipe it.

Content.Server/_Omu/Revs/MoraleEnsured.cs
harryhulse12-alt: Morale system is also predicted - remove both?

CliveOcelot (approved):
for playtest on production as god commands

PR #1761, opened 2026-09-18
[FIX] Reaction Products in the Stomach Properly Clear, Mostly

codeduck101:
## About the PR

reactions occurring inside the stomach will have their products properly cleared out eventually. mostly

## Why / Balance

you're not meant to have things stay in your stomach forever
theoretically this bug can cause people to be unable to ingest things until getting a stomach transplant

## Technical details

added event `SolutionRelayEvent<ReactionAttemptEvent>` to `Content.Shared/Body/Systems/StomachSystem.cs`
on event, stomach will add the outputs of the reaction to the `ReagentDeltas` list so that the reaction's outputs will be appropriately transferred from the stomach to the bloodstream

yes, for multiple small reactions this will cause one billion reagentdeltas, but it seems to handle itself just fine in the end

image
image

very strange thing i have found is that for whatever reason, dying to a ClF3 pill will cause some of the reagent to not be processed. no i do not know why this happens
image

## Media

https://youtu.be/7Vzy-at03xw

## Requirements

## Breaking changes

**Changelog**

:cl: Duuckie
- fix: Reactions occurring in the stomach will no longer have their products remain in the stomach forever, for the most part.  Certain issues may occur with pills.


codeduck101:
hold up let me see if i can fix the deal about ClF3 being cooked

codeduck101:
hmm. i don't think i'm doing this correctly
do hold until i figure something out or someone else gets it

PR #1774, opened 2026-09-20
Make Gorrila gauntlet actualy work, and add emag interaction to it

CliveOcelot:
## About the PR
Ports updated versions of a few systems from wizden (see changes to files without comments for the most part), this now means that the gorilla gauntlet works properly.

As for emag interaction: Normally gorilla gauntlet is intended to unanchor and then launch anything a wrench can unanchor, ex: a lathe. however i added a bit that means that upon emagging, it will try to unanchor whatever it hits, even if normally you wouldn't unanchor it, for example: a wall, or a door. Notably, this only applies to things that can be re-anchored to my knowledge. 

## Why / Balance
Bugfix, alongside it makes it more of a interesting item both for crew and traitors.

## Media
Just spawn a gauntlet, shove anom core in it, then go punch a wall and a lathe, the lathe should move, the wall shouldn't, then emag the gauntlet, and punch a wall and a lathe, both should now move.

## Requirements

## Breaking changes
throwInAir has been removed from TryThrow (it did nothing anyway, as it was defaulted to true, and no system set it to false)
Content.Shared/Throwing/ThrowingSystem.cs is more in line with latest wizden (notably, the entity queries are slightly different, and are commented as such, as currently they way wizden defines them is unsupported on omu.)

**Changelog**
:cl:
- add: Added an emag interaction to the Gorilla gauntlet, once emagged you are now able to launch doors, walls, etc.
- fix: Fixed Gorilla gauntlet not unanchoring objects when it should.


starsoda743:
please add media to the PR when possible

Jerulean (approved):
Tests pass as described in the PR, code checks out as well

